Issue #2, opened 2019-03-27
Partition number checking failing in ww3_ounf


ajhenrique:
A bug in ww3_ounf is causing the netcdf gridded post-processing wrapper to fail. The line  

IF(NBIPART .GT. NOSWLL) THEN

needs to be changed to

 IF(NBIPART .GT. NOSWLL+1) THEN

this reflects the fact NOSWLL sets the number of swell partitions in addition to the windsea partition thus a total of NOSWLL+1 partitions.

@aliabdolali @ajhenrique  @ukmo-ccbunney

ukmo-ccbunney:
This affects the ALLOCATE statement at line 384 in a similar fashion; the TABIPIART array needs to be 1 element larger:

    ALLOCATE(TABIPART(NOSWLL + 1))

ukmo-ccbunney:
Pull request #3 raised.

ajhenrique:
The issue has been solved. To be reintegrated to the NOAA-EMC master pending regtests.

PR #3, opened 2019-03-28
Hotfix: NBIPART max value check failing in WW3_OUNF program 

ukmo-ccbunney:
Reference #2 

When requesting output partitions in _ww3_ounf.ftn_ a check is made to ensure that the number of partitions requested does not exceed that maximum number of partitions generated in the out_grd.ww3 file.

In the current release, this check is sometimes erronously failing as the `NOSWLL` value includes the 0th partition (wind sea), but this is not accounted for in the test. For example, `NOSWLL = 5` means that there are acutally 6 partitions in the out_grid.ww3 file: 5 swell partitions and 1 wind sea.

The value of `NBIPART` needs to be checked against `NOSWLL + 1` to account for this.

This bugfix also includes a small change to the _ww3_tpt1.1_ regtest to move some broken ww3_grid namelist files into a subdirectory. These will be fixed in a future release.


ajhenrique:
Thanks Chris, you won the first puller prize. We will likely merge this to our master after testing. Will keep you posted.

ajhenrique:
I've created an HF_ounfpart branch from the master for this hotfix. The pull request will be merged into that new branch, tested and merged into the master. I'll wait for USACE's hotfix for depth breaking to create a new release tag with both hotfixes, an update to README.md.
@JessicaMeixner-NOAA @aliabdolali

PR #5, opened 2019-04-02
Changes to manual and WWVER to open WW3 v7 development

ajhenrique:
branch open_develop_v7

Adding changes to reference ver 7.00 in manual and WWVER in w3initmd


zhongtouwang2019:
Thank you for your good job

PR #39, opened 2019-05-31
Bug fix to ad3 when running parallel make

ukmo-ccbunney:
When running parallel make, it is possible under certain circumstancs for `ad3` to delete/move compiled MOD files too soon.

This is only an issue if the compiler does not provide a switch to allow a different directory to specified to output MOD files. In this case, the `ad3` program moves the MOD files from the scratch directory to the `mod` directory manually. Howerer, this uses a `*.mod` file glob which can erronously move/delete files from other compilation processes when running a parallel make.

As the compiler may output MOD files with a filename that may be upper/lower/mixed case, it is necessary to use a case insensitive grep to locate the correct MOD file. A `find -iname` works better, but the `-iname` switch to `find` may not be available on all systems.

### DIFF
--- a/model/bin/ad3
+++ b/model/bin/ad3
@@ -303,7 +303,19 @@
   fi
 
   rm -f $path_o/$name.o
-  rm -f *.mod
+  ## Chris B: Dont delete *.mod - could clobber files in parallel make
+  ## Need to delete specific mod file that could be upper/lower/mixed case!
+#  rm -f *.mod ## too indiscriminate
+
+  ## Using find -iname most succinct, but maybe not all version of find have
+  ## the -iname option? 
+#  find . -iname "${name}.mod" -delete  # not all versions of find have -iname?
+
+  # Fall back on grep -i 
+  mods=`ls *.mod | grep -i "${name}.mod 2> /dev/null"`
+  if [ -n ${mods} ]; then
+    rm -f ${mods}
+  fi 
 
 # --------------------------------------------------------------------------- #
 # 3. Compile source code f / f90                                              #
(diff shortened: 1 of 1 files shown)


ajhenrique:
@mickaelaccensi @JessicaMeixner-NOAA please have a look at this pull request and validate.

mickaelaccensi:
Hi Chris,
What do you mean by parallel make ? Are you sure we don't need to also the remove mod files of the subroutines used ? How can we reproduce your error ?

ukmo-ccbunney:
Hi Mickael,
By parallel make, I mean when the "make" command is invoked with the "-j" flag to perform multiple compilations at once. This is now the default when calling w3_make (controlled by the WW3_PARCOMPN environment variable which defaults to 4).

The error can be reproduced by using compiler flags that do not move the "*.mod" files to a different directory (i.e. the mod directory). For instance, for the Cray Fortran Compiler, using "-I $path_m" instead of "-J $path_m" informs the compiler to look in the mod directory, but not place the compiled *.mod files there. In thius case, the ad3 program moves the compiled .mod files to the mod directory. This is the point where the parallel compilation can fail.

You could argue that this is not a bug as using the "-J" compiler flag (in Cray Fortran) fixes the issue. However, the funcitonality exists in ad3 to do this - and it breaks in parallel make if that functionality is used.

So it is a bug - but one that probably no one will encounter if using the correct compiler flags on a modern Fortran compiler!

Cheers,
Chris.

mickaelaccensi:
ok, thanks for the clear exclanation. I agree with this bugfix

ajhenrique:
I am updating the batch queue options for running the regtests matrix at NCEP. Once that is done I'll check the proposed changes and report to reviewers so we can hopefully move ahead with approval of this pull request.

PR #66, opened 2019-06-28
HF slurm regtests

ajhenrique:
Changes in this branch should also be propagated to the master, to allow running the regtests matrix on the NCEP R&D machine, and to provide compilation and runtime metrics in the process.


ajhenrique:
> This has changes to the version number not just updates for slurm regression testing.

Thanks @JessicaMeixner-NOAA, I made an attempt to cherry pick that evidently failed. I'll work on that and write up an approach that will ensure a more successful outcome, and resubmit the pull request.

PR #103, opened 2019-09-26
Corrected th2m and sth2m calculations.

nacbrother:
On lines 1439 and 1440, the calculations for ABX2 and ABY2 were like that of ABX and ABY.  This may have been an oversight when inserted.  Page 112 of the manual lays out the formulas for mean direction and spreading from a2 and b2 coefficients.  To represent cos2Theta and sin2Theta, trigonometric identities were used to exploit the previous calculated values in EC2 and ES2.  The first five moments of spectra and the wave frequency spectrum are useful in representing the energy spectra for every grid point in a model domain reducing space requirements by O(10).  Having these spectra handy allows for access to on-the-fly boundary conditions from reconstructed spectra with minimal loss of accuracy (soon to be shown in a later presentation).

### DIFF
--- a/model/ftn/w3iogomd.ftn
+++ b/model/ftn/w3iogomd.ftn
@@ -1145,6 +1145,8 @@
 !/    01-Mar-2018 : Removed RTD code (now used in post  ( version 6.02 )
 !/                  processing code)
 !/    22-Aug-2018 : Add WBT parameter                   ( version 6.06 )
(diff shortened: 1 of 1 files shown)


ajhenrique:
@nacbrother thank you for your insight. We will be adding your changes to the develop branch soon.

ajhenrique:
Calculation change was confirmed after review by Fabrice Ardhuin. Merge will go ahead.

PR #109, opened 2019-10-06
Add option for reading wind from restart for wmesmf and a few other things

JessicaMeixner-NOAA:
* Added a switch WRST which adds wind to the restart file and then uses it for t0 in wmesmf, which is helpful for when there is no 10m wind from the atmospheric model at initialization. 
* temporary manual updates to remove TEST400 for ST4 until the parameters can be confirmed and updated (possibly should be added as a hot fix to master) 
* removed computed GOTOs (NCO requirement) 
* Added implicit nones to wmesmf 
* Makefile for ESMF has been updated to: 
** have an option for library only 
** updates for hybrid OMP/MPI compilation 

Regtests passed (with the few known exceptions that are known to not be bit for bit)


aliabdolali (approved):

ajhenrique (approved):
This looks good. @JessicaMeixner-NOAA could you please add an issue to remind that this new feature needs to be included in the manual. Thanks for the great work!

JessicaMeixner-NOAA:
@ajhenrique  I'm happy to add that issue.  I already added the WRST switch was added to the list of switches and there is a note about it in the NUOPC section of the manual.  What else would you like me to add? Just more details? What I wrote so far was pretty brief.

ajhenrique:
Hi Jessica

For now that is OK. We can add more descriptive text later. Maybe just create an issue for things that need to be reviewed and expanded in the manual prior to the next public release.

Thanks

Henrique

JessicaMeixner-NOAA:
@ajhenrique  Issue made: https://github.com/NOAA-EMC/WW3/issues/110

ajhenrique:
@JessicaMeixner-NOAA @aliabdolali here are the reports from the regression tests matrix. 
[matrixout.tar.gz](https://github.com/NOAA-EMC/WW3/files/3712425/matrixout.tar.gz)

PR #113, opened 2019-10-10
Fb hybrid switchback

ajhenrique:
Adding back OMPG/OMPH switches for compilation of WW3 using hybrid MPI/OpenMP (NCEP apps).  Adding options for NCEP wcoss Cray compilation that provides b4b consistency.


ajhenrique:
@JessicaMeixner-NOAA please verify is this is what you need for the NCEP NEMSfv3gfs app on the Cray. Also please check the hybrid compilation bits and let me know if I missed something. Thanks!

JessicaMeixner-NOAA (approved):
To clarify, the OMPG and OMPH were never in the develop branch of model/esmf/switch, it was only tested in branches. 

The updates to model/bin/comp.wcoss_cray look good and I think the cmplr.env logic is fine but I didn't test to confirm.

ajhenrique:
> To clarify, the OMPG and OMPH were never in the develop branch of model/esmf/switch, it was only tested in branches.
> 
> The updates to model/bin/comp.wcoss_cray look good and I think the cmplr.env logic is fine but I didn't test to confirm.

Thanks @JessicaMeixner-NOAA!

PR #115, opened 2019-10-22
Fb gefsv12

ajhenrique:
Adding small changes to adjust to NCEP operations:
* Changed compile options for NCEP Dell Phase 3, 
* Removed ampty line on w3iogomd, 
* Expanded output parameters to include Tz, depth, mean direction on ww3_grib, 
* Enforced NOSW=3 on ww3_grid for NCEP operations (NCO switch).

PR #116, opened 2019-10-23
Bugfix in wmesmf flipping nx/ny in exports

JessicaMeixner-NOAA:
The nx/ny for exporting from wmesmf got flipped again.  

The ww3 regtests were not run because they do not test wmesmf.  This was tested in the ufs-weather app and we can now again correctly export z0 from WW3.

### DIFF
--- a/model/ftn/wmesmfmd.ftn
+++ b/model/ftn/wmesmfmd.ftn
@@ -3175,7 +3175,7 @@
 ! -------------------------------------------------------------------- /
 ! 4.  Create ESMF grid with arbitrary domain decomposition to match
 !     the native domain decomposition of the non-excluded points
-!     Note that the native grid layout is dim1=Y, dim2=X
+!     Note that the native grid layout is dim1=X, dim2=Y
 !     Note that coordinates and mask are not needed since this
 !     grid is only used to define fields for a redist operation
 !
@@ -3200,9 +3200,9 @@
             if (ipass.eq.2) then
               ix = mapsf(isea,1)
               iy = mapsf(isea,2)
-              ! native grid layout: dim1=Y, dim2=X
-              arbIndexList(arbIndexCount,1) = iy
-              arbIndexList(arbIndexCount,2) = ix
+              ! native grid layout: dim1=X, dim2=Y
+              arbIndexList(arbIndexCount,1) = ix
+              arbIndexList(arbIndexCount,2) = iy
             endif
           enddo
         endif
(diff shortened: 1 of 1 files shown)


ajhenrique:
@JessicaMeixner-NOAA I've inspected the NOAA-EMC/WW3 repo looking out for the changes  you refer to, at all places including the master, the nx/ny order seems to have remained unchanged (eg, was flipped) since inception of the GitHub repository. Could you provide your last record of when your previous correction might have been made? Do you think there may be other bugfixes that may have not made it in association with this reversal? What do you suggest we can do to enforce testing for this in our current regtests matrix? Thanks.

JessicaMeixner-NOAA:
@ajhenrique  The original fix of this happened here: https://github.com/NOAA-EMC/WW3/commit/cbde39fb21d8fa528dc8653067c10476aab9d053#diff-594a8a5392d3f39ff5e4da6c93770a39 

In the svn->git transition, the history of this file doesn't follow just the trunk/master so I'm really not sure how this got changed, but I think it got changed in a commit or maybe I never actually pushed it to the trunk/master?  Unfortunately I really don't know how it happened to determine if something else might have not made it through.   There is this https://github.com/NOAA-EMC/WW3/blob/master/regtests/bin/run_esmf_test_suite for testing ESMF which you could add to regression testing.  The NEMS regression tests usually are a good way for testing WW3 updates but that is just for importing WW3 variables whereas this is an export of a WW3 variable which is all under development and not a part of any NEMS regression test (yet).

ajhenrique:
Thanks @JessicaMeixner-NOAA. I'll have this PR in alongside PR #115 , both will be required for updating NCEP applications.

ajhenrique:
@JessicaMeixner-NOAA I am merging this to the GEFSv12 branch alongside PR #115. I plan on making this into the implementation branch for GFSv16 and UFS applications in general. We can talk about this option on Monday. For now, this gets the ball rolling for the GEFSv12 tests in preparation for retrospectives this week. PR #115 and #116 will be reviewed by @aliabdolali and @mickaelaccensi prior to merging back to develop this week.

JessicaMeixner-NOAA:
@ajhenrique  I guess it's too late to mention that this is not actually needed for GEFSv12 or GFSv16 as we are not exporting any fields from WW3.  

Do I need to make a new PR so that this can be merged into the develop branch?  

Also, for the S2S UFS application, we will need to point to the develop branch, not the GEFSv12 branch as there are still features to be added.

PR #119, opened 2019-10-29
Fb input interp

RobertoPadilla-NOAA:
This is the branch to interpolate fields from regular grids to a curvilinear grids


ajhenrique:
> In SUBROUTINE W3EXNC the parameter NCIDS should then be declared as:
> INTEGER, INTENT(INOUT) :: NCIDS(NOGRP,NGRPP,NOSWLL + 1)
> The present code assumes NOSWLL == 5:
> 
> https://github.com/NOAA-EMC/WW3/blob/39d1fb73d8ffde828e3070eafc49f6fd5951dfdb/model/ftn/ww3_ounf.ftn#L760

@CarstenHansen does this comment belong here?

ajhenrique (approved):
Changes passed NCEP regtest matrix and comparison of runtime and outputs against file-based input to curvilinear grids. Will require adjustments to account for differences in ice mask reconciliation that will be introduced in a later commit.

PR #129, opened 2019-11-07
manual minor correction

mickaelaccensi:
revert comment on wcc and wcf in manual. see issue


mickaelaccensi:
I don't know why there is the list of the 35 previous commits that have already been integrated in NOAA-EMC develop branch. Anyway, the only file changed is manual/eqs/output.tex

ukmo-ccbunney:
@mickaelaccensi I think the reason why you have all the other commits in your commit history is because you have merged in the develop branch. We had a similar issue here. If you **rebase** rather than **merge** then you just see your own commits (they are rebased onto the tip of the develop branch, rather than merged with a merge-commit).

I _think_ you should be able to clean up your commit history simply by issuing the following command in your branch working copy:

   git rebase develop

You will probably need to do a forced push (git push -f) to get the changes back up to Git Hub without complaints...

Issue #130, opened 2019-11-12
Better handling of error/warning grepping when using cmplr.env


ukmo-ccbunney:
When compiling a particular source file, the compiler erros and warnings are grepped from the file in the "comp" script. When using the `cmplr.env` file to generate the comp/link files, the grep pattern is hardcoded to the simple strings "error" and "warning".

For some compilers this is not specific enough to capture _only_ the specific compiler messages. 

For instance, the cray compiler adds it's own summary message that includes the words "error" and "warning" and these are incorrectly being counted as actual error messages.

Therefore, I propose that the search pattern used in the generated `comp` file can be specified in the `cmplr.env` file. It should default to "error" / "warn" and be overriden where necessary.

ukmo-ccbunney:
Commit 46ff3ef4 details implementation on UKMO bf_gencomp branch.
@mickaelaccensi - do you approve?

mickaelaccensi:
I fully approve !

ukmo-ccbunney:
Thanks @mickaelaccensi !  Could you also cast your expert eye over #124 and let me know if you agree with the changes proposed there too? Thanks.

Issue #124, opened 2019-11-06
OMP flags not generated when using cmplr.env


ukmo-ccbunney:
Reference: issue #19 

I think there is an issue with the way that OMP settings are handled in the `cmplr.env` file.

Currently, OMP compiler flags are only enabled if the `$omp_mod` environment variable is set to "yes", but this is not set anywhere in the file, or in `w3_setup` which uses `cmplr.env` file to produce the `comp` and `link` scripts.

This means that if you set up a comp/link script using `w3_setup` and an entry in `cmplr.env`, it will never contain any OMP flags. `w3_make` will set the `omp_mod` environment variable depending on the program being built, but this will not have any affect as the generated comp/link files do not contain any tests on `$omp_mod`.

Unless I am mistaken, the comp.tmpl and link.tmpl files need to contain a section like the following:
```
if [ "$omp_mod" = 'yes' ]
  then
    opt="$opt $opt_omp"
fi
```

where `$opt_omp` is set in the cmplr.env file. 

The references to `$omp_mod` in cmplr.env should then be removed and the OMP flags should always merged into the comp/link files by `w3_setup`.

ukmo-ccbunney:
See following UKMO/bf_gencomp branch: cb01247
@mickaelaccensi and @ajhenrique - what do you think?

mickaelaccensi:
I agree with your way to handle the OMP flag, it's sounds good to me

Issue #30, opened 2019-05-20
add sea/swell correction options for ww3_uprstr method


ukmo-ansaulter:
Add in  wave spectrum update method, following Lionello et al (1992), to ww3_uprstr.ftn

ukmo-ansaulter:
For assignment to @ukmo-ansaulter

ukmo-ansaulter:
Code has now been updated and tested with Met Office regular and SMC grid trials suites, plus new regtests. Summary of changes is as follows:

- Added UPD5 option; bulk change to wind-sea or full spectrum based on proportion of wind-sea
- Added UPD6 option; as per UPD5 but wind-sea energy distribution is shifted in frequency space using Toba (1973)
- WRST switch is used to read wind data from restart file for UPD5 and UPD6; this is the recommended option, although the facility to read from the anl.XXX file (SWH, WSPD, WDIR) is available
- Regtests have been added for UPD5 and UPD6 options
- The code in ww3_uprstr.ftn has been tidied to make the style for each option a little more consistent

Regtests with the updated code introduce changes in ww3_uprstr.out files; these are correcting a typo for the module name

ukmo-ansaulter:
Review and pull request open on UKMO trusted repo: https://github.com/ukmo-waves/WW3/pull/1

@ukmo-ccbunney assigned for review

ukmo-ansaulter:
UKMO review complete and feature branch merged into UKMO repository develop branch
https://github.com/ukmo-waves/WW3/tree/develop

UKMO pull request to NOAA repository pending other UKMO develop updates

aliabdolali:
@ukmo-ccbunney can we close it?

ukmo-ansaulter:
Yes we can @aliabdolali 👍

PR #132, opened 2019-11-27
UKMO Develop - gencomp and ww3_uprstrt updates

ukmo-ccbunney:
This pull request comprises two sets of commits:
   1. Andy Saulter's updates to ww3_uprstst: adds Lionello/Toba based spectral updates and uses either text file or restart file (WRST switch) based wind inputs alongside SWH and wave spectra (ref issue #30). The `ww3_ta1` regtest group has also been added to the "all" selection of regtests in `matrix.comp`.
   2. Updates to the generic compilation scripts to allow correct handling of OMP switches and grepping of compiler output warning/errors (ref issues #130, #124).

A full regression test matrix has been run (matrixCompFull.out attached):
  * The `mww3_test_03` regtests are not b4b repoducable (known issue).
  * There are some small changes to the `ww3_ta1` regtest outputs, but this is only due to a typo correction in the `ww3_uprstr` screen output (ww3_uprsrt corrected to ww3_uprstrt)
  * The following new regtests have been added to `ww3_ta1`:  UPD6_O, UPD5_U, UPD6_U_cap, 
UPD5_U_cap, UPD5_O, UPD6_U.

[matrixCompFull.out.zip](https://github.com/NOAA-EMC/WW3/files/3896381/matrixCompFull.out.zip)


ajhenrique:
@ukmo-ccbunney I am waiting for the review from @flampouris. I will contact him to check if there was any issue.

flampouris:
@ajhenrique or @aliabdolali could you point me to the run directory of the regression tests?

Thank you

ajhenrique:
> @ajhenrique or @aliabdolali could you point me to the run directory of the regression tests?
> 
> Thank you

Hi @flampouris as per our discussion, the UK Met is a trusted repository, which means they run the complete regtest matrix and we do not replicate the matrix run unless this is necessary for debugging etc. Please contact @ukmo-ccbunney to discuss the details of what you need to complete the review process. Thanks!

model/ftn/ww3_strt.ftn
```diff
@@ -193,6 +193,7 @@
       USE W3WDATMD
       USE W3ODATMD, ONLY: NDSE, NDST, NDSO, NAPROC, IAPROC,           &
                           NAPOUT, NAPERR, FNMPRE
+!/WRST      USE W3IDATMD, ONLY: W3NINP
```
flampouris, line 196: The WRST is used also for ESMF, and overwrites the forcing at the time step of the restart. It is an option for a specific application, not related to the spectra update. 

I consider that these 2 options should not be activated with the same switch, please update.

model/ftn/ww3_uprstr.ftn
```diff
@@ -934,17 +1406,404 @@
...
+!
+! 10. Source code :
+!
+!/
+         REAL, DIMENSION(:), INTENT(INOUT)  :: VATMP
+         INTEGER, DIMENSION(:), INTENT(IN) :: VAMAPWS
+         REAL, INTENT(IN) :: PRCNTG
+         INTEGER, INTENT(IN) :: ISEA1p
+         LOGICAL, INTENT(IN) :: ADJALL
+         CHARACTER(25),PARAMETER :: myname='UPDTWSPEC'
```
flampouris: The name misses an F

flampouris (changes requested):
The source code is clearly written. I can see only one issue:

The WRST is used for updating the forcing too, I consider that it is an option that not all the ww3 users want to exercise while updating the restart file.

model/ftn/ww3_strt.ftn
ukmo-ansaulter: Thanks Stelios

The use of WRST here is deliberate - when running w3uprstr with the UPD5/6 methods we need to read in the background wind speed along with the background wave spectrum. The best way to do this is via the restart file, rather than reading wind in from a separate output file. 

As far as I could tell WRST is the switch that controls wind being in the restart at all, so should be used identically in w3uprstr?? Otherwise I would need to set up the model with two switches, one (WRST) to tell me to have wind in the restart, and a second to tell w3uprstr to read wind from the restart. That seems a bit excessive...

Please let me know if I've missed a point and that WRST also affects the wind in a subsequent cycle in a way that would mess up the usual wind input for a non-coupled model.

model/ftn/ww3_uprstr.ftn
ukmo-ansaulter: Thanks - will get this updated

mickaelaccensi (commented):
my only remarks is for Crayftn (CCE) case in cmplr.env, it was 
  # omp options
  if [ "$omp_mod" = 'yes' ] ; then
    optc="$optc -h omp"
    optl="$optl -h omp"
  else
    # by default, OMP is enabled on Cray ftn:
    optc="$optc -h noomp"

    # ... but don't disable at link stage as is currently needed by SCRIP
    #optl="$optl -h noomp"
  fi

and now it's only replaced by optomp="-h omp"
is it intentional @ukmo-ccbunney ?

model/ftn/ww3_strt.ftn
ajhenrique: @flampouris please reply to @ukmo-ansaulter at your earliest convenience. I would like to have this PR merged to the WW3 develop as soon as possible. Thanks!

ukmo-ccbunney:
> my only remarks is for Crayftn (CCE) case in cmplr.env,
> is it intentional @ukmo-ccbunney ?
Hi @mickaelaccensi,

Yes - there is no easy way of handling that edge case. Plus, the issue really needs to be fixed in the SCRIP code (protective $!/OMP flags around the relevant omp statments/directives in `scrip_remap_conservative.f`). I will raise a Git Issue for this.

It's not an problem for me at the moment because those regtests don't work on the Cray anyhow - I am having to use the GNU compiler. 

Cheers.

ajhenrique:
> It's not an problem for me at the moment because those regtests don't work on the Cray anyhow - I am having to use the GNU compiler.

@ukmo-ccbunney would it be possible to provide a solution for the Cray? We would need that at NCEP.

mickaelaccensi:
my idea for this test was to define to 2 disctinct omp variables for comp and link, something like : 
if [ "$omp_mod" = 'yes' ] ; then
ompc="-h omp"
ompl="-h omp"
else
# by default, OMP is enabled on Cray ftn:
ompc="-h noomp"
ompl=''
fi

ukmo-ccbunney:
> my idea for this test was to define to 2 disctinct omp variables 

@mickaelaccensi, yes - I did consider this, but seemed a bit unnecessary. One would not normally want to have OMP enabled at the compile step, but not the link step (or vice-versa). The only reason I had was as a work around for the SCRIP issue.

However, I am happy to add it if you think it might be useful in other cases though.

Chris.

mickaelaccensi:
i prefer to keep the simple statement optomp="-h omp"

model/ftn/ww3_strt.ftn
flampouris: Hello Andy @ukmo-ansaulter 

With activated the WRST, and for a specific combination of switches, the wind forcing at the time of the restart file is replaced by the wind information included in the restart file, see lines 2245-2255 at wmesmfmd.ftn.

In this case, the **analysis** of the wind field is replaced by a **prediction** of the wind field from an earlier cycle, resulting in discontinuity and potential degradation of the wind forcing.

Therefore, the choices for the update of the wavefield's initial conditions should not have any effect on the forcing.

I hope that this comment clarifies my original review. Please, let me know if there are any additional questions or concerns.

model/ftn/ww3_uprstr.ftn
ukmo-ccbunney: @flampouris - I've corrected the "myname" variable now. Thanks.

model/ftn/ww3_strt.ftn
flampouris: @ukmo-ansaulter check also the PR #146

ajhenrique:
> i prefer to keep the simple statement optomp="-h omp"

@ukmo-ccbunney @mickaelaccensi would this address the -h noomp option?

mickaelaccensi:
actually "-h noomp" option is not needed since the good way to deactivate the omp option is rather to not write OMP statement in the switch file

ajhenrique:
> actually "-h noomp" option is not needed since the good way to deactivate the omp option is rather to not write OMP statement in the switch file

Yes, agreed!

model/ftn/ww3_strt.ftn
ukmo-ansaulter: I see exactly where you are - so WRST implies the actions of both storing wind data in the restart file and giving this wind data precedence over any new wind that uses the same timestamp. In order to get around this we would need to split WRST into two switches, one for each function and then use only the read/write wind in restart part in uprstr.

This doesn't sound like a great idea to me since you would then need to build your ESMF system with two switches to deal with the wind data. Also, from an uprstr point of view, the longer term development of UPD5/6 would want exactly this precedent for the wind to be in place - if we follow Lionello et al's method through completely, then the wind field in the restart should be updated in order to balance the updated wave field's wind-sea component.

As a work around, UPD5/6 already have an option in place where the background wind can be read from the same file as the analysed Hs field, so for those concerned about the precedence associated with WRST the wave model the code can be used in this way without the switch being activated.

Can I suggest therefore, that this is something that is made more explicit in the code and manual documentation rather than making the switch set up more complicated?

ajhenrique: @ukmo-ansaulter I agree. Could you add a runtime warning that informs the user of a potentially unwanted effect when combining switches? @flampouris please comment.

flampouris (commented):
Double post

@ukmo-ansaulter 

- This could be a sufficient comprising solution, I suggest that reviewers from institutes using ESMF should approve the change. 

- My review is focused on the PR's code, we can continue the scientific discussion to a different forum.

model/ftn/ww3_strt.ftn
flampouris: @ukmo-ansaulter

This could be a sufficient comprising solution, I suggest that reviewers from institutes using ESMF should approve the change.

My review is focused on the PR's code, we can continue the scientific discussion to a different forum.

ukmo-ansaulter: Thanks @flampouris - sounds good to me, we will get the changes implemented asap

ajhenrique: @ukmo-ansaulter and @ukmo-ccbunney thank you for your patience. Since requested, this review took 48 days to be initiated, I won't expect that you complete the required revisions in 48 hours, although it seems you have everything lined up to get this done shortly which I greatly appreciate. As a matter of respect to time and effort in our development process, I will discuss with @aliabdolali and other trusted repo code managers ways to certify that our review process follows more reasonable timelines. Thank you.

ukmo-ansaulter: Noting that we took this conversation offline for a short period to clarify the risks and potential solutions for the change I am precis-ing below...

A check through the code indicates that building with the WRST switch carries no risk to the normal operation of the model; any contention is presently limited to the wind data used in coupled systems by ESMF.

The uprstr code, as stands, allows build and operation of the restart update without WRST being a mandatory switch; i.e. ESMF and DA can be operated safely in a consistent build with or without WRST

There is a minor risk of a user deciding to build  a model using the DA module and including WRST without realising the changed impact this would have on ESMF behaviour. There are two potential work arounds for this: 1) add a warning message to the ESMF module and reinforce in manual documentation; 2) update the switch used by ESMF to clobber the T+0 wind file

Based on the above, we have 3 courses of action dependent on preferred timescales and approach:

1. For an immediate merge: Accept PR as it is, including the minor risk that someone chooses to do something daft with the model running a develop branch in the short term. We then set up issues to work on either or both of work arounds 1 and 2 above

2. For a merge early next week: Met Office to update the code and PR with work around 1, assuming the PR would then be acceptable

3. Longer term: Reject the PR and split out the changes. PR with Chris' changes to be submitted late next week. The feature branch for DA to be re-submitted including work around 2) and with review by Jessica/ESMF developers at a later date TBD

This becomes a code managers prerogative now I think, but it is clearly important to get reviewers opinions here. @flampouris can you confirm if options 1 or 2 would be at all acceptable for you? @JessicaMeixner-NOAA could you also provide a view on the risk associated with 1 or 2 (so its not just my say so)?

ajhenrique: @ukmo-ansaulter thanks for summarizing the offline discussions, and thanks all for the great comments and comprehensive evaluation. Given the clearly-stated importance of final reviewer's opinions, which we fully agree, our prerogative is over-ridden and becomes conditional to having those opinions added to this comment thread. @flampouris and @jessica please add your views at your earliest convenience. Thanks!

tjcnrl: Thank-you @ukmo-ansaulter for providing such a helpful summary and thanks to all the reviewers for paying attention to such details.  As is understood by all, the WRST switch does not have any direct conflict with wmesmfmd or running of WW3 under ESMF.  The potential conflict is only that wmesmfmd:GetImport will overwrite the initial wind (wx0,wy0) with what is received from the coupling layer.  What is needed in wmesmfmd:GetImport is a (dynamic) flag to indicate that the initial winds are obtained from the restart and not to be overwritten.  Does such a flag exist in one of the w3*md modules?  If so, then we simply include it in wmesmfmd and add logic to wmesmfmd:GetImport accordingly.  This, I think, can be addressed after this merge has been completed.

ajhenrique: Thank you @tjcnrl for your assessment. Alongside other comments, arguments and evidence provided previously, this answers the only pending concern from @flampouris as folllows:

>     * This could be a sufficient comprising solution, I suggest that reviewers from institutes using ESMF should approve the change.

Issues have been created addressing other concerns that have been identified as enhancements and not essential modifications to this PR.

Once again, thanks to all reviewers for your comprehensive assessment of the proposed changes. I commend in particular reviews that were authentically constructive by going beyond simply identifying issues and risks, but also provided suggested solutions to such issues and risks in a way that fully leverages the strengths of the open development paradigm, towards making ours an improved software package.

ajhenrique (approved):

ajhenrique (approved):

ajhenrique (approved):

aliabdolali (approved):

Issue #1, opened 2019-03-21
Adding a checkpoint restart capability


JessicaMeixner-NOAA:
A requirement for operations, is going to be the ability to restart WW3 from a checkpoint.  This means selecting the correct restart file and then continuing to write output files from that point. 

A proposed solution is to add an option to not only label the restart files with dates such as
restart.YYYYMMDDHHMMSS.ww3 (or some form of YYYYMMDDHH) and to explore the possibility of having output such as out_pnt.ww3 also be for each time (out_pnt.YYYYMMDDHH.ww3) so that you would not have to over-write anything in the binary file or cycle through to find out in initialization where you should be in the binary file. 

We could also consider adding an option to put the output in folders, since this potentially could increase the number of files which is of a concern in coupled models where all models have lots of output files and it could be an issue for the file system. 

@aliabdolali @ajhenrique

ajhenrique:
Planning for the checkpoint feature has started. My initial idea is to keep the restart files are they are for backwards compatibility, but have an option to use the suffix YYYYMMDDHH rather than NNN. A separate checkpoint file (say, snapshot.ww3.YYYYMMDDHH or checkpoint.ww3.YYYYMMDDHH) would be created with an optional flag indicating how many of the latest checkpoint files would be retained as the model runs -- given restart files can be very large, the retain barrier could keep filesystem footprint under control. 

In tandem with the checkpoint file, options to write individual out_grd, out_pnt, partition, and nest outputs would be made available.

Finally, log files would be appended with the continuing run std out if found in the runtime directory.

PR #133, opened 2019-12-02
In line posproc

RobertoPadilla-NOAA:
The additions in this branch give the user the option to have separated output files for out_grd and out_pnt, one file for every time step output provided by the user, but if the user do not want separated files then WW3 will output one big file as for the traditional output.
For this new feature the user has to add 1, 0 or nothing in the user-input files for shel or multi in the line where the initial time , time step and final time is given for grd and pnt output.
1: means the output will be given in separated files, one for every ouptu time step
0: means one file for all output time steps
nothing (not adding anything):The output will be in one file for all time steps.
The additions was coded in such  way that the program will run even if the users do not add anything to the outgrd and outpnt lines for time ouput.


model/ftn/w3iogomd.ftn
```diff
@@ -2604,6 +2615,90 @@
...
+            CO='0'
+          ELSE IF (FCSTTIME .GE. 100) THEN
+            CO=''
+          END IF
+    
+! Convert integer FCSTTIME into a character Hstr with 3 columns 
+          WRITE(Hstr,"(f6.2)") FCSTTIME
+          Hstr(4:4)='p'
+          FCSTNAM='f'//TRIM(ADJUSTL(CO))//TRIM(ADJUSTL(Hstr))
+!!          write(*,*)'w3iogo OutPut Filename: out_grd.', FCSTNAM,'.',FILEXT(:I)
```
JessicaMeixner-NOAA, line 2650: Was there any consideration to putting the valid time instead of a forecast hour as suggested here: https://github.com/NOAA-EMC/WW3/issues/1

JessicaMeixner-NOAA (commented):
Need to add updates to model/inp/* and model/nml/* to explain these new options. 
The other question I have, is if there's a reason to do forecast hour and not valid time of output as part of the name of these new files.  This is at one point what was discussed (https://github.com/NOAA-EMC/WW3/issues/1).

ajhenrique:
Thanks @JessicaMeixner-NOAA. Forecast hour was used after consulting with NCO. However, FV3 will use YYYYMMDD.HHMMSS as a prefix to their checkpointing files, so I may revert this to that format instead of forecast hour. This development will be made on a separate FB_checkpoint branch under my WW3 fork.

JessicaMeixner-NOAA:
I would think that the YYYMMDD.HHMMSS would be a simpler way of keeping track for checkpointing, because if you restarted half way through you would not have to keep track of is FH003 from the check point restart time or a different time.  I think it would require adding a lot more logic to the code versus just writing out the valid time. 

Are there going to be updates to ww3_ou**.ftn for how you process these files or are they just going to processed by taking out_pnt.FH003.glo -> out_pnt.ww3 and then process as before, but just for one time snap?

ajhenrique:
@JessicaMeixner-NOAA I agree. The new approach has been updated on a new branch. Initially the processing will use the .ww3 extension file-by-file. The plan is to port the netcdf and grib2 encoding to become side-by-side so that ww3_ou[t|n][f|p] would not be used.

PR #136, opened 2019-12-31
Fb segfalt grdid

aliabdolali:
@thesser1 @ajhenrique 
Here is Mathieu bug fix for the segmentation fault.
The regression test is done successfully, the comparsion with NOAA develop is attached.
>Insertion of the PDLIB stuff
> Correct unset variable + segmentation fault.

[comparison.zip](https://github.com/NOAA-EMC/WW3/files/4013377/comparison.zip)
AA


ajhenrique:
@aliabdolali as per recent discussion, there are several differences in out_pnt binary files that can't be clearly traced back to the changes in this PR, although it is likely that changing GRDID will change the binary point output. Please confirm if this is the case.

aliabdolali:
> @aliabdolali as per recent discussion, there are several differences in out_pnt binary files that can't be clearly traced back to the changes in this PR, although it is likely that changing GRDID will change the binary point output. Please confirm if this is the case.
@ajhenrique I am doing one test before contacting Matheiu and Aron.
I'll confirm or ask him to take a look.
AA

ajhenrique (approved):
@aliabdolali as per discussion today, looks like all is in good shape.

PR #145, opened 2020-01-14
Gefs v12: Adjusting comp/link options for NCEP wcoss build

ajhenrique:
NEMS has been updated to use generic cmplr.env scripts, there is no need for machine-specific comp/link files for wcoss applications, which also benefits cleaner control of build options across wcoss resources.


aliabdolali (approved):
You are good to go

RobertoPadilla-NOAA (approved):
Everything looks fine.

PR #148, opened 2020-01-15
BF w3_setup

ajhenrique:
Activating omp option in w3_setup and cmplr.env comp/link.tmpl, cleaning up deprecated NCEP R&D machine theia from scripts, eliminating redundant options, correcting activation of -p flag when prof suffix is present


mickaelaccensi (changes requested):
could you revert your changes in w3_setup since we will use Chris solution for omp ?

ajhenrique:
> could you revert your changes in w3_setup since we will use Chris solution for omp ?

This PR has not been merged, I'll close it so the changes it proposes will have no effect on the develop branch.

ajhenrique:
This Pr is embedded in PR #132

PR #161, opened 2020-02-03
Initializing empty array WORDS where an optional parameter, for output fields and points, is stored.

RobertoPadilla-NOAA:
In this bug fix branch a solution was provided for an array (WORDS) that must start as an empty array. That is used to store dates (init date - output time step - end date) for field and point output, it is also used to store an optional user choice to have the output (fields and/or points)  in a separate files, producing one file per output  per-time step or to have all fields (and/or points) in one big file having all fields (and or points) for all output times. That array (WORDS) is used for fields and points output, the array is not initialized as an empty array when is going to read the times and output time step for points, and if it happens that the user provide the optional parameter to have fields output separated but no adding any value for the optional parameter for points, the value provided for fields will be valid for the points output which it will produce separated files for points when it was not the case.

### DIFF
--- a/model/ftn/wminitmd.ftn
+++ b/model/ftn/wminitmd.ftn
@@ -1121,6 +1121,7 @@
         CALL NEXTLN ( COMSTR , MDSI , MDSE2 )
 !
         IF(J .LE. 2) THEN
+          WORDS(1:6)=''
           READ (MDSI,'(A)') LINEIN
           READ(LINEIN,*,iostat=ierr) WORDS
 !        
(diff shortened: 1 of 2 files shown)


JessicaMeixner-NOAA:
This makes reference to "inline post".  Is this a new "inline post" feature or is this again just  a reference to the changes needed to separate the output into per-time step binary files?

RobertoPadilla-NOAA:
The additions and changes in this branch will be added to a new pull request, as the new branch will have code that involves the code in this branch, so we can avoid having conflicts.

PR #166, opened 2020-02-10
Declaring WORDS as empty array and adding call nextline when 2nd line…

RobertoPadilla-NOAA:
In this bug fix branch a solution was provided when the user choose the option to have  a second stream of restart files and adding comments before the second restart line, and also initialize an  array (WORDS) as an empty array. 
A "CALL NEXTLINE()" was added when reading the optional second line for restart file to allow comment lines before reading the second line for restart .
That  WORDS array, this is used to store dates (init date - output time step - end date) for field and point output, it is also used to store an optional user choice to have the output (fields and/or points) in a separate files, producing one file per output per-time step or to have all fields (and/or points) in one big file having all fields (and or points) for all output times. That array (WORDS) is used for fields and points output, the array was not initialized as an empty array when the times and output time step for points are read, and if it happens that the user provide the optional parameter to have fields output separated but no adding any value for the optional parameter for points, the value provided for fields will be valid for the points output which it will produce separated files for points when it was not the case.

### DIFF
--- a/model/ftn/wminitmd.ftn
+++ b/model/ftn/wminitmd.ftn
@@ -1121,6 +1121,7 @@
         CALL NEXTLN ( COMSTR , MDSI , MDSE2 )
 !
         IF(J .LE. 2) THEN
+          WORDS(1:6)=''
           READ (MDSI,'(A)') LINEIN
           READ(LINEIN,*,iostat=ierr) WORDS
 !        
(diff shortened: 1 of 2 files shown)


ajhenrique:
@aliabdolali and @ukmo-ccbunney please review this PR which fixes unwanted behavior when requesting individual files at output times, and a second stream of restart files.

ukmo-ccbunney (approved):
All looks very sensible to me.

ajhenrique:
Changing the landing point for merge from develop to GFS.v16 branch

PR #173, opened 2020-02-29
Bugfix for propagation sequencing to obtain b4b restart runs

ajhenrique:
Bugfix for propagation sequencing to obtain b4b restart runs
* Changes to the behavior of YFIRST in W3PRO[23] for alternating solutions in X and Y propagation directions,
* Changing logic for updating ITLOCH to for truncating intra-spectral propagation solution,
* Adding RWND switch for activating relative winds when current are present for NCEP ESMF applications
* Adding RSTYPE variable for identifying restart type globally


ajhenrique:
Experiments were made at NCEP to verify b4b reproducibility of checkpoint-restart situations. A first run was made with simulation period 0h-4h generating restarts at 1h or 2h for a second run extending from  1h-4h or 2h-4h, respectively. The overlapping simulation periods were compared. Results were b4b in the 2h-4h case, but not in the 1h-4h case. This behavior was tracked to a dependency of W3XYP2/3 on ITIME for setting a flag YFIRST alternating the solution in X and Y directions. A bugfix is proposed setting YFIRST depending on a ratio between current time relative to start time and the global time step DTG. Results were b4b regardless of the restart offset (1h, 2h, 3h etc).

Several runs were made to verify the impact on results, which were found to be minimal, of the order of < 1% for measurable wave heights. An example showing a 120h run comparing Hs from a run with the original YFIRST dependency on ITIME and the new dependency on TIME/DTG are provided in attachment.

Also in attachment, the [results of running the regression test matrix](https://github.com/NOAA-EMC/WW3/files/4271117/matrixComp.zip) are provided (the full data value differences is hosted in a tar bzip2 file: 
ftp://polar.ncep.noaa.gov/pub/waves/reg_matrix/matrixDiff.tbz

As expected, there are many differences in regtests. Inspection of the differences reveal that these correspond to small discrepancies in values of integral spectral parameters, caused by differences in the sequencing pattern that mostly occur in multigrid cases ([w3wavemd is called a second time for output only at the end of each global integration step within the wmwavemd](https://github.com/ajhenrique/WW3/blob/FB_yfirst_rst/model/ftn/wmwavemd.ftn#L1019), which makes ITIME be incremented +1 without computations being performed. The latter behavior created a disruption in the alternating pattern when using the multigrid driver.

* 120-day simulations
** Hs from develop:

** Hs from new branch:
WW3Modif
** Differences in Hs develop-new branch:
WW3Ori-Mod

ajhenrique:
@ukmo-ccbunney @CarstenHansen in the light of previous reviews, please assess this pull request which is required for NCEP applications. Thanks!

model/esmf/switch
```diff
@@ -22,6 +22,7 @@ MLIM
 TR0
 BS0
 XX0
+RWND
```
JessicaMeixner-NOAA, line 25: Do you want this for GEFS or just GFS (will this slow down GEFS)?

JessicaMeixner-NOAA (approved):

ukmo-ccbunney (commented):
@ajhenrique The logic in w3proXmd and w3wavemd look sensible. Your explanation of the differences in the multi-grid regtests makes sense, but can you explain why there are differences in the single grid tests? From what I can see, the logic in w3pro2md and w3pro3md should result in the same order of X/Y propagation as the original method.

ajhenrique:
> From what I can see, the logic in w3pro2md and w3pro3md should result in the same order of X/Y propagation as the original method.

@ukmo-ccbunney great question. The differences are due to the alternating split in spectral space. I may have incorrectly coded that part. The original
`ITLOCH = ( NTLOC + 1 - MOD(ITIME,2) ) / 2`
became
`ITLOCH = ( NTLOC + MOD(NINT(DTTEST/DTG),2) ) / 2`
but should have been
`ITLOCH = ( NTLOC + 1 - MOD(NINT(DTTEST/DTG),2) ) / 2`
I will make the correction and rerun the single grid regression tests to verify this provides b4b results relative to develop. Good eye, thanks!

ukmo-ccbunney:
> 
> 
> > From what I can see, the logic in w3pro2md and w3pro3md should result in the same order of X/Y propagation as the original method.
> 
> @ukmo-ccbunney great question. The differences are due to the alternating split in spectral space. I may have incorrectly coded that part. The original
> `ITLOCH = ( NTLOC + 1 - MOD(ITIME,2) ) / 2`
> became
> `ITLOCH = ( NTLOC + MOD(NINT(DTTEST/DTG),2) ) / 2`
> but should have been
> `ITLOCH = ( NTLOC + 1 - MOD(NINT(DTTEST/DTG),2) ) / 2`
> I will make the correction and rerun the single grid regression tests to verify this provides b4b results relative to develop. Good eye, thanks!

@ajhenrique Yes - I was a bit suspicious of this line at when I first saw it. However, considering that ITIME starts at 1, I concluded that you had coded it correctly. Unless ITIME is zero at the first time that statement is reached... Perhaps that is the case? 

EDIT: Actually, yes that is probably the culprit line. I was forgetting that the first timestep will be at +DTG, so the statement `MOD(NINT(DTTEST/DTG),2)` should be equal to 1 for the first timestep, meaning that your fix should hopefully replicate the X/Y order.

@ukmo-ccbunney indeed I just confirmed the change makes the logic identical to the original one. We'll rerun the regtests to make sure the single-grid cases are b4b.

ajhenrique:
> @ajhenrique Yes - I was a bit suspicious of this line at when I first saw it. However, considering that ITIME starts at 1, I concluded that you had coded it correctly. Unless ITIME is zero at the first time that statement is reached... Perhaps that is the case?
> 
> EDIT: Actually, yes that is probably the culprit line. I was forgetting that the first timestep will be at +DTG, so the statement `MOD(NINT(DTTEST/DTG),2)` should be equal to 1 for the first timestep, meaning that your fix should hopefully replicate the X/Y order.

@ukmo-ccbunney I found that by making this change there were still very small differences in the listed non-identical single-grid cases. I added some write statements and found out that ITIME in many cases is set in a way that it is not consistent with other runs, so that the alternating pattern becomes out of phase between the old and the new approach. I did not dig too deep into the reason why ITIME is so erratic, I'll create an issue to investigate that. Bottom line is that the new dependency for alternating X/Y or determining the spectral splitting is more robust that the ITIME dependency, is consistent across the board, and provides b4b results in restart situations. Let me know if that warrants approval or you would like to look further into the ITIME behavior. Thanks!

aliabdolali (approved):

ukmo-ccbunney (approved):
@ajhenrique I agree that the new method of selecting the X/Y propagation order is an improvement over the existing method and am happy that the regtest differences can be accounted for. Thumbs up from me!

PR #197, opened 2020-05-11
Bugfix/spconv

mickaelaccensi:
update matrix.base
update headers


aliabdolali:
@mickaelaccensi thanks for the update.

PR #203, opened 2020-05-21
Final changes for GEFSv12

ajhenrique:
Integrating final changes for GEFSv12 back to develop


ajhenrique:
These changes do not affect regression tests and are exclusively for NCEP operations. Extensive testing has been made for the target application. Please advise if regtest matrix is needed for this case.

ajhenrique:
Please disregard these changes made at error.

PR #211, opened 2020-06-15
Reintegrate Production/GFS.v16 to develop

ajhenrique:
Add recent changes to WW3 for use in NCEP's GFS.v16 upgrade.
* Add OMP loop in wmgridmd for optimizing initialization time
* Remove the excessive esmf field gather in wmesmfmd.ftn 
* Add allocatable variables to WMUPDV in call to interpolation sbr
* bugfix for WORDS array
* Adding options for building WW3 within ESMF/NEMS in the Orion
* Adjusting if statement in w3_make to avoid crash during build is netcdf library has non-numerical version label


JessicaMeixner-NOAA (changes requested):
The description of this PR is inaccurate as the only update that has not gone back into the develop branch from production/GFS.v16 is the last bullet point "Adjusting if statement in w3_make to avoid crash during build is netcdf library has non-numerical version label" 

Also, the develop branch was merged into the GFS.v16 branch which I don't think should have happened as GFS.v16 code should be in a close to frozen state, if not frozen, I thought? So does that need to be reverted as well?

Issue #242, opened 2020-08-12
Zero windspeed on zero meridian


SanderHulst:
The _averaging_ code in WAVEWATCH appears not to properly project 0.125 degree ECMWF surface wind data on our global 30 minute grid, while 0.25 degree GFS data seems to be projected to the target grid just fine. An important aspect to keep in mind is that both grids have their left origin on the zero meridian. Would they have been on -180, the problem would manifest itself there.

A bug is present that affected both datasources but only results in the zero windspeed in the ECMWF case because of a threshold in the same averaging code.

It was traced to wmupdmd.ftn:
```
          DO J=IXS0, IXSN
              JJ=J
              IF ( FLAGLL ) THEN
                IF ( ICLOSE.NE.ICLOSE_NONE ) JJ = 1 + MOD(J-1,NXI)
                IF ( JJ.LT.1 .OR. JJ.GT. NXI ) CYCLE
<compute weight of source grid cell>
           END DO 
           IF ( NXA(IX,0)  .GT. 0 ) THEN
              RT     = SUM ( RXA(IX,:) )
              IF ( RT .LT. 0.7 ) THEN <---- NOTE THE THRESHOLD
                NXA(IX,:) = 0
                RXA(IX,:) = 0.
              END IF
            END IF
          END DO
```
J is the index of the source grid cell considered. The MOD is used to fold the right to the left side of the source grid, but it does not fold the left to the right because negative indices pass the mod statement as is. In the case of a negative index or zero, the CYCLE will cause the source point not to be considered. Below the output of a [test program](https://github.com/NOAA-EMC/WW3/files/5062046/test_fortran_stuff.txt) is printed that simulates 0.25 degree source data and a half a degree target.

```
X      : 0.000
SX     : 0.500
SXI    : 0.250
NXI    :  1440
MXA    :     4
IX     :   361
XFL    : -.250
XFR    : 0.250
IXC    :     1
IXS0   :    -2
IXSN   :     4
Skipped JJ    -2
Skipped JJ    -1
Skipped JJ     0
JJ, TOGGLED, XFL, XFR, XSL, XSR, R1   :    1   0   -0.250    0.250   -0.12500    0.12500    0.500
JJ, TOGGLED, XFL, XFR, XSL, XSR, R1   :    2   0   -0.250    0.250    0.12500    0.25000    0.250
[...]

PR #243, opened 2020-08-12
No data on zero meridian when resolution ratio > 2

SanderHulst:
Please see NOAA-EMC/WW3#242 for a full description.

Oh, and release/zero_windspeed_on_zero_meridian should have been called feature/zero_windspeed_on_zero_meridian.


aliabdolali:
I checked the regression on NOAA HPC, with the diff in known non b4b cases.
[MatrixDiffNCEP.zip](https://github.com/NOAA-EMC/WW3/files/6143721/MatrixDiffNCEP.zip)

JessicaMeixner-NOAA:
@SanderHulst  thanks for your patience with my slow review.  I've been looking at this off and on for the past few days and I'm still stuck on the following: 

Why do we not also want to apply the same theory on https://github.com/NOAA-EMC/WW3/blob/develop/model/ftn/wmupdtmd.ftn#L1369-L1372 
so that we would then have: 
```
              IF ( FLAGLL .AND. ICLOSE.NE.ICLOSE_NONE ) THEN
                J1 = 1 + MOD(J1-1+NXI,NXI)
                J2 = 1 + MOD(J2-1+NXI,NXI)
              END IF
```
?

Is it just that your case you had the >2, so you focused on that, or is it that the ratio of difference is smaller so even without the wrapped interpolation of the points you don't see the line like you do in the case you showed in the issue.

SanderHulst:
I did not consider the equal resolution case, but I don't think it will show the same behavior. The root cause is a negative index that is not folded. The code for MXA .NE 2 is:
```
            IF ( FLAGLL ) THEN
              IXC    = 1 + NINT ( MOD (                             &
                          1080.+X0+REAL(IX-1)*SX-X0I , 360. ) / SXI )
              IXS0   = IXC - 1 - MXA/2
              IXSN   = IXC + 1 + MXA/2
            ELSE
              IXC    = NINT ( 1. + ( X0+REAL(IX-1)*SX - X0I ) / SXI )
              IXS0   = MAX (  1  , IXC - 1 - MXA/2 )
              IXSN   = MIN ( NXI , IXC + 1 + MXA/2 )
            END IF
            DO J=IXS0, IXSN
              JJ=J
              IF ( FLAGLL ) THEN
                IF ( ICLOSE.NE.ICLOSE_NONE ) JJ = 1 + MOD(J-1+NXI,NXI)
                IF ( JJ.LT.1 .OR. JJ.GT. NXI ) CYCLE

```
The actual problem is in the do J=IXSO, IXSN loop that looks left and right of the cell center and causes J to be negative .

In the equal resolution case, J1 can not be negative because XR always lies between 1 and NX. Right?

JessicaMeixner-NOAA:
@SanderHulst apologies on the super slow response.  I just took a look at this again, and you are correct we don't need anything for the other case.  I think this is good as is.  Thanks again for this update!

JessicaMeixner-NOAA (approved):

PR #279, opened 2020-12-12
removing file that should not be in repository

JessicaMeixner-NOAA:


mickaelaccensi:
we should even add a rule in gitconfig to not take into account .swp files

JessicaMeixner-NOAA:
@mickaelaccensi we have in .gitignore: 

```
*.swp
*/*.swp
*/*/*.swp
*/*/*/*.swp

```
already. I would have thought that would have prevented this, but apparently not?

mickaelaccensi:
it's because it's a hidden file. So we need to add : 
```
.*.swp
*/.*.swp
*/*/.*.swp
*/*/*/.*.swp
```

JessicaMeixner-NOAA:
@mickaelaccensi added!

ukmo-ccbunney:
Do we need all the directory permutations?
I think that the rule `*.swp` should match any file in the repo, even if it is in a directory.

On my system `*.swp` ignores all .swp files, regardless of directory depth or if the filename is hidden (starts with a dot).

JessicaMeixner-NOAA:
I didn't think you had to have the directory permutations but I'm not sure?  Maybe we just already had that file in the repo before the *.swp was added?

mickaelaccensi:
it's not the case on my system. The dot must be added to filter hidden files.

ukmo-ccbunney:
> 
> 
> it's not the case on my system. The dot must be added to filter hidden files.

Interesting that it is different on your system - maybe a different version of Git (I'm on 2.24.3).
So, perhaps we just need:
```
*.swp
.*.swp
```
??

JessicaMeixner-NOAA:
Combined into  PR #284

aliabdolali:
@JessicaMeixner-NOAA It is merged with https://github.com/NOAA-EMC/WW3/pull/284, can we delete it?

PR #295, opened 2021-01-12
Reduce initialization time for GFSv16

JessicaMeixner-NOAA:
This PR is for the production/GFS.v16 branch.  It comments out advertising of export fields from WW3 and the "SetExport" routine in addition to commenting out possible input fields for connection except u10/v10 which is used.   This simplifies the connector initialization in NUOPC and significantly reduces the overall init time for GFSv16.   This update should NOT go back into the develop branch as we want these fields advertised.  In addition updates/optimization in later versions of ESMF make these simplifications unnecessary.

### DIFF
--- a/model/ftn/wmesmfmd.ftn
+++ b/model/ftn/wmesmfmd.ftn
@@ -1799,18 +1799,18 @@
 !
 ! 2.e Set export fields
 !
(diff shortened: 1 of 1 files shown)


aliabdolali (approved):
It looks ok.

Issue #307, opened 2021-02-04
Add threadprivate for saved variables 


JessicaMeixner-NOAA:
In debugging a case where the UFS coupled model was not thread safe, it was determined that WW3 was the problem.  OMP standard says: 
Local variables with the SAVE attribute declared in procedures called from a parallel region are implicitly SHARED. The specification will be clarified.
and adding in w3srce: 
!/OMPG/!$omp threadprivate( TAUNUX, TAUNUY)
solved the issue reported in the UFS https://github.com/ufs-community/ufs-weather-model/issues/367 

To be safe, all local saved variables that could be used threaded, need to have the thread private added.  

Thanks @JamesAbeles-NOAA for finding the solution to this issue.

JessicaMeixner-NOAA:
@thesser1 @aliabdolali I did not address PDLIB variables in PR #315 for hopes that someone with more experience with PDLIB would be better at determining which of the save variables should have similar thread private declarations.

JessicaMeixner-NOAA:
@thesser1 @aliabdolali  -- I'm going to close this issue as it's my understanding PDLIB/unstructured does not have threading capabilities yet.  Please reopen this issue or a new issue should this be closed in error.

Issue #306, opened 2021-02-04
Add Gaea for comp/link for UFS


JessicaMeixner-NOAA:
Add compiling info for Gaea for use in the UFS.

JessicaMeixner-NOAA:
In adding other compilers for UFS, I simplified the name of the structures to be "hera.intel" "Gaea.intel" etc... @aliabdolali @awest-noaa will this be an issue for you?  See: https://github.com/JessicaMeixner-NOAA/WW3/tree/feature/portww3forufs for proposed changes.  This requires a minor cmake list update on the ufs-weather-model side as well.

JessicaMeixner-NOAA:
The changes that were needed at the ufs-model level to the cmake list can be seen here: https://github.com/JessicaMeixner-NOAA/ufs-weather-model/commit/7a2d7d2b021d7ee72e6097d0db9ec3ea39ab04c7

@aliabdolali @awest-noaa I hope to make a PR with these changes by the end of this week.  Please let me know if this will not be acceptable for coastal applications

PR #284, opened 2021-01-05
removing comp.* link.* files as we now use the cmplr.env

JessicaMeixner-NOAA:
This PR removes all of the model/bin/comp.<> & link.<> files as we have transitioned to using the cmplr.env files. 

I have not run regression tests yet, as I wanted to make sure there were no objections to the removal of these files.  If someone is still using one of the independent comp/link files, we could add them to the cmplr.env. 

Update: 
PR #283 which removes unused files in the bin directory and PR #279 which got rid of .swp files not intended to be committed to repo  have been combined with this PR.

Issue #214, opened 2020-06-23
regtest: OMP, number of tasks and restart reproducibility


JessicaMeixner-NOAA:
Add regtests to test that different number of OMP threads and different number of tasks produce the same answer.  Also add a restart reproducibility test (ie. if you write a restart half way through a run, and start from that restart you should get the same answer).

aliabdolali:
Hi @JessicaMeixner-NOAA 
I have a setup for restart file reproducibility for the tp2.17 (unstructured) which might need revisiting. For example, can we add commands in matrix and copy restart001.ww3 to another directory? Also we need two ww3_shel.inp for two consecutive runs. 
if yes, it is very simple to set it.
i.e. cp ww3_tp2.17/work_a/restart002.ww3 ww3_tp2.17/work_b/restart.ww3
run_test ...
or the structured grids, you pick one and tell me the name, I can do the same.
AA

JessicaMeixner-NOAA:
@aliabdolali awesome.  We might do a couple of the structured grid tests?  Something that looks more like GEFSv12 or GFSv16 grid set-up with a multi run and then also a single grid? (Although single grid will likely be reproducible if a multi-grid case is). I'll have to look and see which one would be best.

MatthewMasarik-NOAA:
@JessicaMeixner-NOAA or @aliabdolali, I believe we have run_tests to check each of these:  OMP, tasks, and restart reproducibility.  Do any of those need more attention, or could this issue be closed?

PR #315, opened 2021-02-18
Updates to make WW3 thread safe 

JessicaMeixner-NOAA:
This adds thread private declarations for saved variables (Issue #307) for non-PDLIB cases.   There is a long outstanding issue #214 to add regression tests for testing OMP threads and this PR does not address this long standing need.  There will be a regression test added for the UFS system for which this bug was originally found.  While only the TAUNX/Y in w3srce needed to be added to solve that case, the principal for why that was the fix was applied to the variables I found (except PDLIB).   Thanks again to @JamesAbeles-NOAA for finding the solution! 

Additionally in this PR two files were removed that should have been removed in PR #284 but were missed when combining a few PRs together. 

Lastly, this PR also included porting #306 for WW3 to be used in UFS applications. 

Only the usual regtests had differences.  
[matrixout20200218.tar.gz](https://github.com/NOAA-EMC/WW3/files/6005022/matrixout20200218.tar.gz)


JamesAbeles-NOAA (approved):

JessicaMeixner-NOAA:
As a sanity check, I ran the WW3 regression tests with OMP, OMPH tests with a different number of threads and the only regression test that failed is known to  fail: ww3_tp2.10/./work_MPI_OMPH  -- granted I tried this also with the develop branch and got the same result.

ukmo-ccbunney (commented):
Hi @JessicaMeixner-NOAA 
These changes all look very sensible.
I will run some relevant regression tests next week.

ukmo-ccbunney (approved):
Hi @JessicaMeixner-NOAA ,
I've run the regression tests for the hybrid MPI-OMP tests on our system and the results are all identical with those from the develop branch. I'm happy to sign this one off now.

aliabdolali (approved):

Issue #206, opened 2020-06-03
Extend parameter set for coupled modelling 


ukmo-ansaulter:
Enhancement to extend output parameter set and restart file storage options required for coupled atmosphere-wave-ocean modelling. The extensions are based on Met Office testing with UM-WWIII-NEMO models coupled via OASIS

ukmo-juan-castillo:
Some changes in the code are required to allow the coupling of some extra wave variables to atmosphere and ocean models, just by adding them to the list of fields available for coupling. Furthermore, a new series of variables required for coupling need to be calculated in the source terms, such as the wind stress transferred to the ocean. All the fields that are in the 'send' coupling list need to be written in the wave restart file, if the source terms are not calculated before the first time step of the model. This is required in order for the coupling fields sent to other models to contain appropriate values at the first time step, when coupling always takes place, in order to preserve the continuity of a 'cycling' run.

ukmo-ansaulter:
@ukmo-juan-castillo will be the main developer for these changes

@JessicaMeixner-NOAA these changes will include some work very similar to your WRST changes, so hopefully OK to check in with you for some advice/review as we go along?

JessicaMeixner-NOAA:
of course @ukmo-ansaulter  

I have a feeling we might need more variables for restart reproducibility in the ESMF coupling as well for variables not calculated until after the first time step.

ukmo-juan-castillo:
> of course @ukmo-ansaulter
> 
> I have a feeling we might need more variables for restart reproducibility in the ESMF coupling as well for variables not calculated until after the first time step.

This has already been tackled in issue #232, and now it is possible to include a list of variables in the restart file via a namelist variable. At the moment only variables that can be exchanged via OASIS coupling can be written to the restart, but modifying the code to add new fields is straightforward.

PR #327, opened 2021-03-05
UKMO Staging jan2021

### LINKED ISSUE #207: Optimisations for coupled modelling performance (OASIS)
Enhancements to improve performance and MPI communications for coupled atmosphere-wave-ocean modelling. Changes based on Met Office work using a UM-WWIII-NEMO modelling framework coupled via OASIS

### LINKED ISSUE #290: ww3_multi hangs on when creating restart file with IOSTYP=2 or =3
the bug appears when using IOSTYP 2 or 3, it works well with IOSTYP=1

By tracking where it keeps waiting, it seems to be for some processors in w3wavemd :
CALL MPI_WAITALL due to positive value of NRQSG2

and for the dedicated output processor in w3iorsmd :
```
!/MPI                        CALL MPI_WAITALL                         &
!/MPI                           ( 1, IRQRSS(IB), STAT1, IERR_MPI )
```

I'll look for a regtest that highlights the bug

### LINKED ISSUE #314: Interpolation of input fields in ww3_prnc
While testing a development branch (see #285) we discovered some problems when reading from input an air density field. We checked the input file and it is OK, but in the model output some zeroes appear in the middle of the domain, and strange interpolation is made along the land/sea boundary. Two main problems were found:

1. For points in the input field adjacent to land, interpolation still takes place with land values, which are set to zero by default. In consequence, the grid interpolation that takes place in ww3_prnc generates non-physical values. This can be fixed by checking if there are adjacent land points and recalculating the normalization factors ignoring these land points. An improvement is possible by calculating weights in land points adjacent to ocean instead of setting them to zero, so that interpolation takes place with valid values. This could be useful for atmosphere fields where in principle there are valid values in the whole grid domain, but the modifications required are large.

2. When the boundary cells read from the nesting file are wrong and are not ocean grid points (as it is the case in the ww3_tp2.15 regtests), the weights are still used and are all zero, so the final interpolated value is zero. This can be avoided by just not interpolating this type of points and keeping the original value. It is not ideal but it will avoid having spurious values at the land/sea boundaries.

It is quite possible that this fix will change results of some tests, as it affects how the forcing fields are read.

ukmo-ccbunney:
The staging branch comprises two PRs from the ukmo-waves repository:
  * https://github.com/ukmo-waves/WW3/pull/18 : Bug fix for ww3_multi hanging issue
  * https://github.com/ukmo-waves/WW3/pull/19 : Bug fix for interpolation weights calculation in ww3_prnc.

Fixes #290 
Fixes #207 
Fixes #314 

Full regression test output for each PR are available in the PRs links above.

Regression test outputs for this combined staging branch will be provided below.


model/ftn/w3iorsmd.ftn
```diff
@@ -609,7 +604,7 @@
 !/MPI                          IF ( IP .EQ. NAPRST ) THEN
 !/MPI                              WRITEBUFF(1:NSPEC) = VA(1:NSPEC,JSEA)
 !/MPI                            ELSE
-!/MPI                              JSEA   = JSEA - 2*((IB-1)/2)*RSBLKS
+!/MPI                              JSEA = 1 + MOD(IB-1,2)*RSBLKS
```
JessicaMeixner-NOAA: Do the bug fixes here warrant changing the date for this file?

JessicaMeixner-NOAA:
Does this close any issues? I want to say there is at least one issue this PR would close.

model/ftn/w3iorsmd.ftn
ukmo-ccbunney: I was going to say no, as it as it doesn't actually change the format of the file... but I suppose it could change how an earlier file was read in w.r.t. JSEA? @ukmo-juan-castillo - can you confirm?

ukmo-ccbunney:
> 
> 
> Does this close any issues? I want to say there is at least one issue this PR would close.

@JessicaMeixner-NOAA : yes - I've edited the description to link in the relevant issues.

ukmo-ccbunney:
@ukmo-juan-castillo  and I are happy with the regtest results for this PR now. The output of the full regression test maxtrix compared against `develop` is here: 

[matrixComp_staging_jan2021.zip](https://github.com/NOAA-EMC/WW3/files/6156335/matrixComp_staging_jan2021.zip)

Summary of differences relavent to this PR:
  - `mww3_test_04`: As expected only the ‘d’ test where we changed the test to write the restart change results
 - `ww3_tp2.14`: Differences in `r-toy.nc` and `r-ww3.nc` are caused OASIS, appending to files rather than overwriting them when a test is re-run; reproducibility is something I want to investigate later with Jessica
 - `ww3_tp15`: as expected due to corrections in the interpolation of netcdf input fields

Also, the usual culprits which have known differences
- `ww3_tp2.18`: Existing known differences in output to screen
- `mww3_test_03`: Known to be not b4b.

Over to you @aliabdolali  :)

aliabdolali:
Hi @ukmo-juan-castillo and @ukmo-ccbunney 
I am going to check it at our end today.

aliabdolali:
Hi @ukmo-juan-castillo and @ukmo-ccbunney
The output of the full regression test maxtrix compared against develop is here:
[MatrixDiffNCEP.zip](https://github.com/NOAA-EMC/WW3/files/6164850/MatrixDiffNCEP.zip)
I have similar differences as you mentioned. tp2.14 are identical on our HPC and the ww3_tp2.10 has always differences on our HPC. The rests are the same.

ukmo-ccbunney:
> 
> 
> Hi @ukmo-juan-castillo and @ukmo-ccbunney
> The output of the full regression test maxtrix compared against develop is here:
> [MatrixDiffNCEP.zip](https://github.com/NOAA-EMC/WW3/files/6164850/MatrixDiffNCEP.zip)
> I have similar differences as you mentioned. tp2.14 are identical on our HPC and the ww3_tp2.10 has always differences on our HPC. The rests are the same.

@aliabdolali I notice that you have some differences in the `mod_def.*` files for **mww3_test_07**. I wasn't expecting to see any differences there (and we have not in our tests). Do you have any idea what is happening there (it must be some insignificant difference as all the other output files are identical for that test)?

Everything else looks as expected.

aliabdolali:
> > Hi @ukmo-juan-castillo and @ukmo-ccbunney
> > The output of the full regression test maxtrix compared against develop is here:
> > [MatrixDiffNCEP.zip](https://github.com/NOAA-EMC/WW3/files/6164850/MatrixDiffNCEP.zip)
> > I have similar differences as you mentioned. tp2.14 are identical on our HPC and the ww3_tp2.10 has always differences on our HPC. The rests are the same.
> 
> @aliabdolali I notice that you have some differences in the `mod_def.*` files for **mww3_test_07**. I wasn't expecting to see any differences there (and we have not in our tests). Do you have any idea what is happening there (it must be some insignificant difference as all the other output files are identical for that test)?
> 
> Everything else looks as expected.

@ukmo-ccbunney I checked another PR and I had similar non-identical results in `mod_def.*` files for **mww3_test_07**. So it is intel/our HPC related.

ukmo-ccbunney:
>> @ukmo-ccbunney I checked another PR and I had similar non-identical results in mod_def.* files for mww3_test_07. So it is intel/our HPC related.

Phew - that's a relief. Also - this is interesting behaviour as sometimes I also have differences in the `mod_def.ww3` files for one particular  regtest (I forget which one now). I think it might be a compiler issue as the differences usually go away if the filesystem path to the WW3 installation directory has the same number of characters (yes - bizarre I know!!); hence I think it is a GNU Fortran bug/oddity*


\* I sometimes forget that compilers are written by humans too!

aliabdolali (approved):

JessicaMeixner-NOAA:
@ukmo-ccbunney I remember we had this happen before because there was an uninitialized variable being stored in the mod_def.ww3 file.  I'm not sure if it's the same file/test or not, but something like that could explain why some compilers it's fine and others it is not.

JessicaMeixner-NOAA (approved):

ukmo-ccbunney:
Fixes #206

PR #328, opened 2021-03-09
Fb matrix divider

aliabdolali:
Hi @thesser1 @ukmo-ccbunney @mickaelaccensi 
Following our discussion on the slow speed of our regression suite, I modified the script I used to divide the main matrix into a given number of subsets (it is defined by defining the number of tests in each subset), and tested it with the most recent code. The entire regtest took 3-4 hr in 7 subsets (each 100 test) on 24 cores. It would definitely faster if we change n=24 to n=8 cores. 
There is still room for improvement, but I wanted to provide an engineering solution without touching any part of the code. In this way, we just duplicate the model directory and let each individual matrix does its simulation separately from the same parent and store all the solution in one place.

The extra step after matrix file generation is defining the max number of tests in each subset and executing ./bin/matrix_divider.sh 
@JessicaMeixner-NOAA, what do you think?

@thesser1 @ukmo-ccbunney @mickaelaccensi could you check it at your end and provide suggestions?
One thing that I'll add in the future is dividing the parallel vs serial tests, so we can optimize the number of tests in each subset better. It will be an easy implementation.


JessicaMeixner-NOAA:
@aliabdolali thanks for providing an engineering solution.  I'd be happy to check this out.  Can you provide usage instructions?

aliabdolali:
> @aliabdolali thanks for providing an engineering solution. I'd be happy to check this out. Can you provide usage instructions?

In line 37, define _maxlist_. The default is 100 which means 7 seubsets for the whole ~700 tests.
Once you generated the matrix, execute this script and it splits the matrix, and submit the jobs.

JessicaMeixner-NOAA:
Thanks @aliabdolali and sorry you originally included the instructions and I missed it.

regtests/bin/matrix_divider.sh
```diff
@@ -0,0 +1,78 @@
...
+#Replace matrix.out > matrix?.out, model > model?
+  cp -r ../model ../model$i
+  sed -i 's/'matrix.out'/'matrix${i}.out'/gI' matrix$i
+  sed -i 's/'model'/'model${i}'/gI' matrix$i
+  echo "  echo ' '"                                                                     >> matrix$i
+  echo "  echo '     **************************************************************'"   >> matrix$i
+  echo "  echo '     *  end of WAVEWATCH III matrix$i of regression tests         *'"   >> matrix$i
+  echo "  echo '     **************************************************************'"   >> matrix$i
+  echo "  echo ' '"                                                                     >> matrix$i
+sbatch matrix$i
```
JessicaMeixner-NOAA: I would remove this as this would be machine dependent and this could then not be used by others who are not using slurm or if we port to another machine that does not use sbatch.

regtests/bin/matrix_divider.sh
```diff
@@ -0,0 +1,78 @@
...
+#    Copyright 2013 National Weather Service (NWS),                           #
+#       National Oceanic and Atmospheric Administration.  All rights          #
+#       reserved.  WAVEWATCH III is a trademark of the NWS.                   #
+#       No unauthorized use without permission.                               #
+#                                                                             #
+# --------------------------------------------------------------------------- #
+# --------------------------------------------------------------------------- #
+# 1.  clean up and definitions                                                #
+# --------------------------------------------------------------------------- #
+rm before
```
JessicaMeixner-NOAA: Perhaps also add this to the end for cleaning up?   Do we want to add these files to .gitignore?

regtests/bin/matrix_divider.sh
aliabdolali: Sure, I'll remove it.
Once other thing that I'll do for ourself is adding another script which does cloning the FB and Dev, download the tar from ftp, prepare matrix, divide it and submit. So I will include this to that.

regtests/bin/matrix_divider.sh
aliabdolali: Sure,
Yes, we should add  them to .gitignore

JessicaMeixner-NOAA:
This is a great engineering fix and it's really nice that it's automated.  We should see if this also works for others.  I like the idea of committing this first, which would test that this split produces the same as the non-split matrix, but I still think changing from 24->8 is needed.       

Something that I don't like about this is that it copies the model folder which then you lose git capabilities because these files aren't then tracked.  That aspect of this is not my favorite, but without a major build change this definitely simplifies the running of the regression tests in multiple parts and that is really nice. I also like the fact that all the regression tests are already in the regrets folder.  When I do this manually I usually just use multiple clones. 

A possible future improvement would be to look at how we are splitting them instead of just equally into parts.  For example, serial jobs could be separated and appropriate resources could be requested.  Or other splitting for knowing which tests take what amount of time.

aliabdolali:
> This is a great engineering fix and it's really nice that it's automated. We should see if this also works for others. I like the idea of committing this first, which would test that this split produces the same as the non-split matrix, but I still think changing from 24->8 is needed.
> 
> Something that I don't like about this is that it copies the model folder which then you lose git capabilities because these files aren't then tracked. That aspect of this is not my favorite, but without a major build change this definitely simplifies the running of the regression tests in multiple parts and that is really nice. I also like the fact that all the regression tests are already in the regrets folder. When I do this manually I usually just use multiple clones.
> 
> A possible future improvement would be to look at how we are splitting them instead of just equally into parts. For example, serial jobs could be separated and appropriate resources could be requested. Or other splitting for knowing which tests take what amount of time.

Later, I will add something to check if the job is done, then all extra model? should be deleted. I'll modify it now to reflect your suggestions and ask you to review, after the mege, we will ask others to check at their ends. 
Your last suggestion was sth I thought about it, but that requires further work. I am working on it ...

JessicaMeixner-NOAA:
There's no need to merge before others check, they can check before it is merged.

aliabdolali:
@JessicaMeixner-NOAA Do you want me to change 24 to 8 in this FB?

JessicaMeixner-NOAA:
No, because the regression tests results will change and this should be a clean comparison for this commit.

aliabdolali:
@JessicaMeixner-NOAA @ukmo-ccbunney @thesser1 @mickaelaccensi 
I made some further progress
I added another script that separates serial and parallel tests, then we can define the number of tests in each subset for parallel (_maxlist1_) and serial(_maxlist2_) matrix. 
I also made it compatible with ifremer, ukmet and erdc infrastructures, but it should be tested on these platforms to make sure it is working and then we need to optimize _maxlist1_ and _maxlist2_.
please see
**[regtests/bin/matrix_divider_p.sh](https://github.com/aliabdolali/WW3/blob/fb_matrix_divider/regtests/bin/matrix_divider_p.sh)** 
We also can allocate one core for serial jobs.

ukmo-ccbunney:
Hi @aliabdolali 
This looks like a good idea. I will give it a test.

I have one small issue when running the script though - the path to `../model` is not correct if I run the script in the `regtests/bin` directory. Where do you intend the script to be run? I assumed it would be run in `matrix/bin` as that is where the matrix files are...

To get it to run I had to copy my `matrix` file from `regtests/bin` to `regtests`.

Perhaps you always have your matrix files in the `regtests` dir?

ukmo-ccbunney:
>> Something that I don't like about this is that it copies the model folder which then you lose git capabilities because these files aren't then tracked. That aspect of this is not my favorite, but without a major build change this definitely simplifies the running of the regression tests in multiple parts and that is really nice. I also like the fact that all the regression tests are already in the regrets folder. When I do this manually I usually just use multiple clones.

@JessicaMeixner-NOAA - something that I have been working on (but is not working properly yet) is having a "build" directory defined in WW3 that contains the `obj`, `mod` and `scratch` directories. That way, each invocation of w3_make can specify its own unique build directory, if required. It's a work in progress at the moment as I need to update the make_makefile.sh part to ensure the paths are set correctly in there. I will also need to figure out a way of handling changes to the `switch` file.

Perhaps this might tie in with Ali's changes at a later date (to avoid copying the whole model directory?)

regtests/bin/matrix_divider_p.sh
```diff
@@ -0,0 +1,117 @@
...
+  cat list_mpi_$i >> matrix$count
+  cp -r ../model ../model$count
+  sed -i 's/'matrix.out'/'matrix${count}.out'/gI' matrix$count
+  sed -i 's/'model'/'model${count}'/gI' matrix$count
+  echo "  echo ' '"                                                                     >> matrix$count
+  echo "  echo '     **************************************************************'"   >> matrix$count
+  echo "  echo '     *  end of WAVEWATCH III matrix$count of regression tests     *'"   >> matrix$count
+  echo "  echo '     **************************************************************'"   >> matrix$count
+  echo "  echo ' '"                                                                     >> matrix$count
+  echo "rm ../model$count"                                                              >> matrix$count
```
ukmo-ccbunney: needs to be "rm -r" to remove directory

regtests/bin/matrix_divider_p.sh
```diff
@@ -0,0 +1,117 @@
...
+  cat list_serial_$i >> matrix$count
+  cp -r ../model ../model$count
+  sed -i 's/'matrix.out'/'matrix${count}.out'/gI' matrix$count
+  sed -i 's/'model'/'model${count}'/gI' matrix$count
+  echo "  echo ' '"                                                                     >> matrix$count
+  echo "  echo '     **************************************************************'"   >> matrix$count
+  echo "  echo '     *  end of WAVEWATCH III matrix$count of regression tests     *'"   >> matrix$count
+  echo "  echo '     **************************************************************'"   >> matrix$count
+  echo "  echo ' '"                                                                     >> matrix$count
+  echo "rm ../model$count"                                                              >> matrix$count
```
ukmo-ccbunney: needs to be "rm -r" to remove directory

regtests/bin/matrix_divider_p.sh
```diff
@@ -0,0 +1,117 @@
...
+  cat list_2.14 >> matrix$count
+  cp -r ../model ../model$count
+  sed -i 's/'matrix.out'/'matrix${count}.out'/gI' matrix$count
+  sed -i 's/'model'/'model${count}'/gI' matrix$count
+  echo "  echo ' '"                                                                     >> matrix$count
+  echo "  echo '     **************************************************************'"   >> matrix$count
+  echo "  echo '     *  end of WAVEWATCH III matrix$count of regression tests     *'"   >> matrix$count
+  echo "  echo '     **************************************************************'"   >> matrix$count
+  echo "  echo ' '"                                                                     >> matrix$count
+  echo "rm ../model$count"                                                              >> matrix$count
```
ukmo-ccbunney: needs to be "rm -r" to remove directory

ukmo-ccbunney (changes requested):
HI @aliabdolali - I hope you don't mind me hijacking the review.
I've spotted a few "rm" commands that need a "-r" flag to remove the model* directories.

regtests/bin/matrix_divider_p.sh
aliabdolali: Done

regtests/bin/matrix_divider_p.sh
aliabdolali: done

regtests/bin/matrix_divider_p.sh
aliabdolali: done

aliabdolali:
> Hi @aliabdolali
> This looks like a good idea. I will give it a test.
> 
> I have one small issue when running the script though - the path to `../model` is not correct if I run the script in the `regtests/bin` directory. Where do you intend the script to be run? I assumed it would be run in `matrix/bin` as that is where the matrix files are...
> 
> To get it to run I had to copy my `matrix` file from `regtests/bin` to `regtests`.
> 
> Perhaps you always have your matrix files in the `regtests` dir?



> Hi @aliabdolali
> This looks like a good idea. I will give it a test.
> 
> I have one small issue when running the script though - the path to `../model` is not correct if I run the script in the `regtests/bin` directory. Where do you intend the script to be run? I assumed it would be run in `matrix/bin` as that is where the matrix files are...
> 
> To get it to run I had to copy my `matrix` file from `regtests/bin` to `regtests`.
> 
> Perhaps you always have your matrix files in the `regtests` dir?

@ukmo-ccbunney Where do you run matrix_ukmet? I assume you do it in regtests. Then, do the same for matrix_divider_p.sh as matrix is located there.

aliabdolali:
> > > Something that I don't like about this is that it copies the model folder which then you lose git capabilities because these files aren't then tracked. That aspect of this is not my favorite, but without a major build change this definitely simplifies the running of the regression tests in multiple parts and that is really nice. I also like the fact that all the regression tests are already in the regrets folder. When I do this manually I usually just use multiple clones.
> 
> @JessicaMeixner-NOAA - something that I have been working on (but is not working properly yet) is having a "build" directory defined in WW3 that contains the `obj`, `mod` and `scratch` directories. That way, each invocation of w3_make can specify its own unique build directory, if required. It's a work in progress at the moment as I need to update the make_makefile.sh part to ensure the paths are set correctly in there. I will also need to figure out a way of handling changes to the `switch` file.
> 
> Perhaps this might tie in with Ali's changes at a later date (to avoid copying the whole model directory?)

@ukmo-ccbunney this would be great. We can tackle it later.
BTW, I made a few other changes, please try it and if you like it, we can do the final retouch and merge it. 
I would remove matrix_divider.sh and will keep matrix_divider_p.sh. It is well organized as it separates serial and parallel tests.

regtests/bin/matrix_divider_p.sh
```diff
@@ -0,0 +1,118 @@
...
+  cat list_mpi_$i >> matrix$count
+  cp -r ../model ../model$count
+  sed -i 's/'matrix.out'/'matrix${count}.out'/gI' matrix$count
+  sed -i 's/'model'/'model${count}'/gI' matrix$count
+  echo "  echo ' '"                                                                     >> matrix$count
+  echo "  echo '     **************************************************************'"   >> matrix$count
+  echo "  echo '     *  end of WAVEWATCH III matrix$count of regression tests     *'"   >> matrix$count
+  echo "  echo '     **************************************************************'"   >> matrix$count
+  echo "  echo ' '"                                                                     >> matrix$count
+  echo "rm -r ../model$count"                                                              >> matrix$count
```
JessicaMeixner-NOAA: Instead of removing/cleaning the extra model directories in the top level script, which usually comes up as an error and since we remove the extra model directory in this script, maybe in the matrix? script, you check if the model? dir exists, if it does, remove and copy a new model? from model and then remove at the end of the script.    That way if for whatever reason you have to re-run the matrix? You don't have to re-run the top level script?

aliabdolali: @JessicaMeixner-NOAA Done, thanks for the suggestion. I went one step further and also put an if statement to remove matrix? if any.

aliabdolali: @JessicaMeixner-NOAA could you refer to the line you still like a change?

regtests/bin/matrix_divider_p.sh
```diff
@@ -0,0 +1,123 @@
...
+#serial jobs
+ for i in `seq -f '%02g' 0 "$((matrixno2 - 1))"`; do
+#echo $i
+#Replace matrix.out > matrix?.out, model > model?
+  (( count = count + 1 ))
+  if [ -d "${HOME}/model${count}" ]; then rm -Rf ${HOME}/model${count}; fi
+  if [ -f "matrix${count}" ]; then rm -f matrix${count}; fi
+  cat before >> matrix$count
+  cat list_serial_$i >> matrix$count
+  cp -r ${HOME}/model ${HOME}/model$count
```
JessicaMeixner-NOAA: This line should be added to the matrix$count script.  If it is not, you cannot re-run matrix$count as you remove model$count within that scripts (see line 110).

aliabdolali: @JessicaMeixner-NOAA Done

JessicaMeixner-NOAA: I just ran the script, and am getting a little bit of weird output: 

```
if [ -d ../model77 ]; then rm -Rf ../model77; fi
cp -r ../model7 ../model77 
...
./bin/run_test -b slurm -c hera.intel -S -T -s PR2_UQ -w work_PR2_UQ_g -m grdset_g ../model7 mww3_test_04
...
rm -r /scratch2/NCEPDEV/climate/Jessica.Meixner/PR_WW3/PR328/WW3/model7

```

Shouldn't it be something like: 
```
if [ -d ../model7 ]; then rm -rf ../model7; fi
cp -r ../model ../model7 
...
./bin/run_test -b slurm -c hera.intel -S -T -s PR2_UQ -w work_PR2_UQ_g -m grdset_g ../model7 mww3_test_04
...
rm -rf /scratch2/NCEPDEV/climate/Jessica.Meixner/PR_WW3/PR328/WW3/model7

```

aliabdolali: @JessicaMeixner-NOAA fixed

JessicaMeixner-NOAA: Thanks @aliabdolali  this is now working!

JessicaMeixner-NOAA:
@aliabdolali thanks for addressing all my comments!

aliabdolali:
@JessicaMeixner-NOAA 
I ran Develop without breaking it into subsets and I used this PR in 11 subsets. There is no missing case when we subset it (as expected)
Here are the outcomes:
[MatrixDiff.zip](https://github.com/NOAA-EMC/WW3/files/6197837/MatrixDiff.zip)
I also reran Develop twice to make sure the non-identical cases are not because of dividing the matrix. 
It seems it is ready to be merged.

JessicaMeixner-NOAA (approved):
Thanks @aliabdolali

PR #329, opened 2021-03-10
Update issue templates

aliabdolali:


.github/ISSUE_TEMPLATE/add-development-request.md
```diff
@@ -0,0 +1,10 @@
+---
+name: Add Development Request
```
JessicaMeixner-NOAA, line 2: I think this all looks fine, but I'm curious if there is a way to preview how this will actually look?

aliabdolali: @JessicaMeixner-NOAA On the top right, click on **...** and select **view file**

JessicaMeixner-NOAA: Screen Shot 2021-03-17 at 9 34 03 PM
So this is what this will look like?

JessicaMeixner-NOAA (approved):
It's hard to imagine how this will look or interface even with the viewer.  Maybe this is something that we add and then adjust later?  

Another class of issues might be documentation?

.github/ISSUE_TEMPLATE/add-development-request.md
aliabdolali: I think so. As you suggested, we can merge and modify if further changes are required.

aliabdolali:
> It's hard to imagine how this will look or interface even with the viewer. Maybe this is something that we add and then adjust later?
> 
> Another class of issues might be documentation?

I am going to add documentation issue tab

Issue #345, opened 2021-03-31
grib encoding issues for PolarStereo grid


JessicaMeixner-NOAA:
For the polar stenographic grid in ww3_grib when GDTN = 20 
KGDS( 1) should be 5 not 6 

KGDS( 1) = 5  means Earth assumed represented by WGS84 see https://www.nco.ncep.noaa.gov/pmb/docs/grib2/grib2_doc/grib2_table3-2.shtml

JessicaMeixner-NOAA:
Still need to merge this update to the develop branch.

aliabdolali:
@JessicaMeixner-NOAA, it requires grib regtests, so the one I made for GFSv16 has this already. Do you think it should be on a separated PR?
i can make grib regtest totally separate + this fix.

JessicaMeixner-NOAA:
Awesome, glad this is include in that PR. That PR would close this issue then.

PR #346, opened 2021-03-31
Fix for arctic grib files 

JessicaMeixner-NOAA:
# Pull Request Summary
Update way polar stenographic grid is encoded in ww3_grib 

## Description
When GDTN = 20 in ww3_grib for polar sterographic grids, we should use the "Earth assumed represented by WGS84" option which requires changing KGDS( 1) = 6 to setting KGDS( 1) = 5. 

### Issue(s) addressed
* This will fix the issue reported by the Alaska region for the Aleutian Islands appearing shifted. 
* See issue #345 for more details. 

### Testing
* ww3_grib was tested on WCOSS via v16 global-workflow 

Fix made with help from:  @aliabdolali @arunchawla-NOAA @BoiVuong-NOAA @GeorgeGayno-NOAA

### DIFF
--- a/model/ftn/ww3_grib.ftn
+++ b/model/ftn/ww3_grib.ftn
@@ -620,6 +620,9 @@
 !       (17) Projection center flag
 !       (18) Scanning mode
 !
+!  Projection for PolarStereo grid was changed from 
+!  KGDS( 1) = 6 to KGDS( 1) = 5 (Earth assumed represented by WGS84)
+!/NCEP2        KGDS( 1) = 5
 !/NCEP2        X0 = MOD(XGRD(1,1) + 360.,360.)
 !/NCEP2        XN = MOD(XGRD(NY,NX) + 360., 360.)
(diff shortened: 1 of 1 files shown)


model/ftn/ww3_grib.ftn
```diff
@@ -620,6 +620,9 @@
 !       (17) Projection center flag
 !       (18) Scanning mode
 !
+!  Projection for PolarStereo grid was changed from
```
aliabdolali, line 623: It can refer to Octet No Table 3.2
! Projection for PolarStereo grid was changed from KGDS( 1) = 6 to KGDS( 1) = 5 (Earth assumed represented by WGS84 - Octet No 15 Table 3.2)

aliabdolali (approved):
Other than that minor comment, it is ok

model/ftn/ww3_grib.ftn
JessicaMeixner-NOAA: better?

aliabdolali: awesome

PR #388, opened 2021-05-27
Remove obsolete OMPX switch 

### LINKED ISSUE #371: Remove OMPX switch
It appears that the OMPX switch is no longer used.  For example see comment in code here: 
https://github.com/NOAA-EMC/WW3/blob/develop/model/ftn/w3wavemd.ftn#L83-L84

However, it is still in many switch files and in the makefile.  Let's clean it out of the whole code as it is not used.  Please let me know if the comment in the code is incorrect.  If no objection is made, the plan will be to remove the switch in late May 2021.

JessicaMeixner-NOAA:
# Pull Request Summary
Remove OMPX switch Issue #371

## Description
The OMPX switch has been obsolete for a while, this removes it from the makefile system as well as the manual. 

### Issue(s) addressed
* closes #371

### Testing
* matrix_ncep was run using hera.intel and all tests ran without error.  The expectation is no answer changes as this switch already was not used, except for the normal not b4b.  The regression tests need to be re-run before being merged to the develop branch.  The matrix summary will be added at that time.


aliabdolali:
I will review this today.

JessicaMeixner-NOAA:
Thanks @aliabdolali I just submitted the regtests

aliabdolali (approved):
OMPX is mentioned in a header of model/ftn/w3wavemd.ftn:84, so I am not sure if we want to keep it as it is the history or clean it. other than that, I am happy with the changes.

JessicaMeixner-NOAA:
> OMPX is mentioned in a header of model/ftn/w3wavemd.ftn:84, so I am not sure if we want to keep it as it is the history or clean it. other than that, I am happy with the changes.

The comment is: 
```
!/    15-Sep-2020 : Bugfix FIELD allocation. Remove     ( version 7.11 )
!/                  defunct OMPX switches.
```
so I think its valid to keep it there.  Unless we change our strategy of keeping the history, I don't think comments like this should be deleted.   If anything we should be adding "removed x switch" in the other routines as we are doing this.

JessicaMeixner-NOAA:
[matrixCompFull.txt](https://github.com/NOAA-EMC/WW3/files/6641075/matrixCompFull.txt)
[matrixCompSummary.txt](https://github.com/NOAA-EMC/WW3/files/6641076/matrixCompSummary.txt)
[matrixDiff.txt](https://github.com/NOAA-EMC/WW3/files/6641077/matrixDiff.txt)

Only the normal not-b4b tests are different

PR #396, opened 2021-06-01
Bring CMEPS mediator compatibility to WW3 cap

uturuncoglu:
# Pull Request Summary
This PR aims to bring capability to run WW3 wave component through the use of CMEPS mediator. The initial implementation is done as a part of HAFS application and supports ATM-WAV and ATM-OCN-WAV (with HYCOM) configurations.

## Description
This PR adds following,
* The corner coordinates is added to WW3 grid in the ESMF/NUOPC cap. This is required for CMEPS since ESMF Mesh is used to represent model grids/meshes and ESMF Mesh generation routine requires corner coordinates. This is also required for conservative regridding.
*  `mediator_present = .true.` needs to be added to `WAV_attributes` to use CMEPS mediator.
* Since WW3 and FV3 domains does not overlap entirely in the regional HAFS application. The data coming from CMEPS is merged with the data read from wind forcing file (`wind.ww3`). The special subroutine in the cap reads the forcing file and merge with the data coming from coupling to provide complete set of wind field to WW3. The merge mask is also generated on the fly by checking fill value of the fields coming from CMEPS mediator. This is also controlled by extra attribute `merge_import = .true.` in `WAV_attributes` and it is not required for the global application.
* The full regression test suit in the UFS Weather model does not indicate any answer change and existing WW3 coupled configurations that uses NUOPC connector rather than CMEPS runs without any issue and answer changes.

### Issue(s) addressed
* Is there an issue associated with this development (bug fix, enhancement, new feature)?
WW3 can be used with CMEPS mediator to exchange data among the components.

### Check list  
* Is your feature branch up to date with the authoritative repository (NOAA/develop)? Yes
* Make sure you have checked the [checklist for a developer submitting to develop](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-a-developer-submitting-to-develop), [checklist for a developer submitting to develop](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-a-developer-submitting-to-develop) and [updating version number](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-updating-version-number)
* Reviewers: @JessicaMeixner-NOAA

### Testing
Full UFS Weather Model RTs were run including existing configurations with wave coupling 
There is no change in the regression tests
* If a new feature was added, was a new regression test added?
N/A
* Have regression tests been run?
Yes, full regression tests were run without any issue on Orion and Hera.
* Which compiler / HPC you used to run the regression tests in the PR? 
* Please provide the summary output of matrix.comp (_matrix.Diff.out_, _matrixCompFull.out_ and _matrixCompSummary.out_):    
Please indicate the expected changes in the outputs ([excluding the known list of non-identical tests](https://github.com/NOAA-EMC/WW3/wiki/How-to-use-matrix.comp-to-compare-regtests-with-master#4-look-at-results)).


model/ftn/wmesmfmd.ftn
```diff
@@ -534,6 +554,20 @@
       if (ESMF_LogFoundError(rc, PASSTHRU)) return
 !
 ! -------------------------------------------------------------------- /
+! Check if coupled with CMEPS mediator or not
```
JessicaMeixner-NOAA, line 560: Is this just to check if the CMEPS mediator exists or if any mediator exists?

uturuncoglu: This check is just for keeping CMEPS modifications as isolated. By this way, WW3 cap can be used with/without CMPES and we could keep continue to support NUOPC connector type coupling.

JessicaMeixner-NOAA: I know this was tested and updated for the CMEPS mediator specifically, but is this checking to see if there is any mediator or just the CMEPS mediator?  The comment implies to me that it's the CMEPS mediator but the code seems to be general for any mediator.

JessicaMeixner-NOAA:
@uturuncoglu can you update the manual (this is the relevant part: https://github.com/NOAA-EMC/WW3/blob/develop/manual/app/nuopc.tex) or at least create a new issue so that we can track that these updates need to be added to the manual at some point?

model/ftn/wmesmfmd.ftn
uturuncoglu: It is just for CMEPS. This specific attribute was already used by CMEPS and I decided to use it to follow the conventions.

uturuncoglu:
@JessicaMeixner-NOAA I think that creating issue about the documentation is better idea since we did not finalize the code and it could change it slightly based on the test of the global application but if you want me to add section to the documentation that is also fine for me.

uturuncoglu:
@JessicaMeixner-NOAA I created an issue about the documentation. Here is the link https://github.com/NOAA-EMC/WW3/issues/399

aliabdolali:
I ran the full regtest on hera with expected known non-b4b tests
```
mww3_test_03
mww3_test_07
ww3_tp2.10
ww3_tp2.15
ww3_tp2.18
```
[CMEPS.zip](https://github.com/NOAA-EMC/WW3/files/6625111/CMEPS.zip)

JessicaMeixner-NOAA:
The regression tests that will matter most for this is to test the ufs-weather-model.  However, we should probably wait for the WW3 PR that updates the code before running those.  

@danrosen25 @BinLiu-NOAA anyone else want to review this code?   I have done my first quick review and regression tests for previous ufs-weather-model passed, but they should be reviewed and I should probably do a more in-depth review as well.

aliabdolali:
@JessicaMeixner-NOAA I think we should associate a milestone to this development on github and define the version accordingly.

JessicaMeixner-NOAA:
@aliabdolali sounds good.  

@danrosen25 @BinLiu-NOAA @uturuncoglu @aliabdolali I'm going to set up a meeting to do a group review of this code.

aliabdolali:
Here is the milestone
https://github.com/NOAA-EMC/WW3/milestone/25

model/ftn/wmesmfmd.ftn
```diff
@@ -271,6 +277,19 @@
       integer               :: natGridID
       logical               :: natGridIsLocal
       type(ESMF_RouteHandle):: n2eRH
+!
+! --- Mediator
+      logical        :: med_present = .false.
+      character(256) :: flds_scalar_name = ''
+      integer        :: flds_scalar_num = 0
+      integer        :: flds_scalar_index_nx = 0
```
danrosen25, line 288: flds_scalar_index_nx and flds_scalar_index_nx are domain metadata that allows CMEPS to convert a mesh back to 2d space for diagnostic outputs.

model/ftn/wmesmfmd.ftn
```diff
@@ -656,13 +692,62 @@
...
+          call ESMF_LogWrite(trim(cname)//': flds_scalar_name = '// &
+            trim(flds_scalar_name), ESMF_LOGMSG_INFO, rc=rc)
+          if (ESMF_LogFoundError(rc, PASSTHRU)) return
+        end if
+
+        call NUOPC_CompAttributeGet(gcomp, name="ScalarFieldCount", &
+          value=cvalue, isPresent=isPresent, isSet=isSet, rc=rc)
+        if (ESMF_LogFoundError(rc, PASSTHRU)) return
+        if (isPresent .and. isSet) then
+          read(cvalue, *) flds_scalar_num
```
danrosen25: My preference for converting strings is to use the ESMF conversion utility (i.e. ESMF_UtilString2Int)
https://earthsystemmodeling.org/docs/release/ESMF_8_1_1/ESMF_refdoc/node6.html#SECTION060144800000000000000

model/ftn/wmesmfmd.ftn
```diff
@@ -789,6 +874,27 @@
       call ESMF_UtilIOUnitGet(idse); open(unit=idse, status='scratch');
       close(idsi); close(idso); close(idss); close(idst); close(idse);
 !
+! 1.g Get merging option for regional coupling that domians does not
+! overlap complately. This will blend the data coming from forcing with
+! the data coming from coupling.
+!
+      if (med_present) then
+        call NUOPC_CompAttributeGet(gcomp, name="merge_import", &
```
danrosen25: This forces "merge_import" feature to only be available if med_present = .true.

model/ftn/wmesmfmd.ftn
```diff
@@ -656,13 +692,62 @@
...
+            trim(flds_scalar_name), ESMF_LOGMSG_INFO, rc=rc)
+          if (ESMF_LogFoundError(rc, PASSTHRU)) return
+        end if
+
+        call NUOPC_CompAttributeGet(gcomp, name="ScalarFieldCount", &
+          value=cvalue, isPresent=isPresent, isSet=isSet, rc=rc)
+        if (ESMF_LogFoundError(rc, PASSTHRU)) return
+        if (isPresent .and. isSet) then
+          read(cvalue, *) flds_scalar_num
+          write(logmsg,*) flds_scalar_num
```
danrosen25: Make sure to follow log writing conventions established by WW3.

model/ftn/wmesmfmd.ftn
```diff
@@ -1750,6 +1916,9 @@
 ! -------------------------------------------------------------------- /
 ! 1.  Check that required import fields show correct time stamp
 !
+      if (med_present) then
+      allUpdated = .true.
```
danrosen25, line 1930: This by-passes the import data-dependency. The import data-dependency restarts winds from an external component. During CMEPS restart runs this DataInitialize subroutine should check imported data time stamp to determine if CMEPS has restarted winds.

model/ftn/wmesmfmd.ftn
```diff
@@ -2826,6 +3049,44 @@
         enddo
       endif
 !
+! 2.h Set ESMF import grid corner coordinates, only for coupled with mediator
+!
+      if (med_present) then
```
danrosen25: Instead of using med_preset for adding corners we want to check the grid/mesh type then add corners as needed. This can always be on.

JessicaMeixner-NOAA: can this be for even if a mediator is not present?

model/ftn/wmesmfmd.ftn
```diff
@@ -2826,6 +3049,44 @@
         enddo
       endif
 !
+! 2.h Set ESMF import grid corner coordinates, only for coupled with mediator
+!
+      if (med_present) then 
+        call ESMF_GridAddCoord( impGrid, &
```
JessicaMeixner-NOAA: Check to make sure we can't get to this section of code w a unstructured grid and we need to have a if GTYPE if statement

model/ftn/wmesmfmd.ftn
```diff
@@ -4626,6 +4933,135 @@
...
+      endif
+!
+! -------------------------------------------------------------------- /
+! Create merging mask
+!
+      if ( impGridIsLocal ) then
+        do j = elb(2),eub(2)
+          do i = elb(1),eub(1)
+            mmsk(i,j) = 0.0
+            if (dptr(i,j).lt.fillVal) then
```
danrosen25, line 5035: Using less than catches more possible fill values but do we want to catch more possible fill values? Ideally fillVal = NaN on systems where this is available because a NaN will propagate if used in interpolation.

model/ftn/wmesmfmd.ftn
```diff
@@ -6551,6 +6986,191 @@
...
+      rewind(mdsf)
+
+      ! read header information
+      ! this was inside of w3fldo call but since we are opening file
+      ! once and rewinding, the header need to be read
+      read(mdsf, iostat=ierr) tsstr, tsfld, nxt, nyt, &
+        gtypet, filler(1:2), tideflag
+
+      ! read input
+      call w3fldg('READ', idfld, mdsf, mdst, mdse, nx, ny, &
```
JessicaMeixner-NOAA, line 7110: look into if there is time interpolation

model/ftn/wmesmfmd.ftn
```diff
@@ -6551,6 +6986,191 @@
...
+      subroutine ReadFromFile (idfld, fldwx, fldwy, time0, timen, rc)
+!/
+!/                  +-----------------------------------+
+!/                  | WAVEWATCH III           NOAA/NCEP |
+!/                  |           U. Turuncoglu           |
+!/                  |                        FORTRAN 90 |
+!/                  | Last update :         18-May-2021 |
+!/                  +-----------------------------------+
+!/
+!/    DD-MMM-YYYY : Origination.                        ( version 7.XX )
```
JessicaMeixner-NOAA: you can go ahead and fill in DD-MM-YYYY with today or other date (maybe the 18-May 2021)?

model/ftn/wmesmfmd.ftn
JessicaMeixner-NOAA: Should we add this description as a comment into the code?

model/ftn/wmesmfmd.ftn
JessicaMeixner-NOAA: Should we go ahead and update to use the new function or make an issue to do this in the future?

model/ftn/wmesmfmd.ftn
JessicaMeixner-NOAA: These write statements are fine, should these be written every time or should they be behind: 

if (verbosity.gt.0) 

statements for the log messages that are not for every run and more just if you want the extra information?

model/ftn/wmesmfmd.ftn
JessicaMeixner-NOAA: Ideally, this would not be for a "med_present" only case.  Can we try to run without this if statement? I can help with at least ensuring that it doesn't break other features on the global side if needed.  At the least, we should make an issue to remove this.

model/ftn/wmesmfmd.ftn
JessicaMeixner-NOAA: I checked and we only get to this routine for the select GTYPES, so no need for a if GTYPE statement: 

 ```
     if ( (GTYPE.eq.RLGTYPE).or.(GTYPE.eq.CLGTYPE) ) then
         write(msg,'(a)') trim(cname)// &
           ': Creating import grid for Reg/Curvilinear Mode'
         call ESMF_LogWrite(trim(msg), ESMF_LOGMSG_INFO)
         call CreateImpGrid( gcomp, rc=rc )
         if (ESMF_LogFoundError(rc, PASSTHRU)) return
      elseif (GTYPE.eq.UNGTYPE) then
         write(msg,'(a)') trim(cname)// &
           ': Creating import mesh for Unstructured Mode'
         call ESMF_LogWrite(trim(msg), ESMF_LOGMSG_INFO)
         call CreateImpMesh( gcomp, rc=rc )
```

File: model/ftn/wmesmfmd.ftn (hunk 7 above)
Comment (JessicaMeixner-NOAA) [reply model/ftn/wmesmfmd.ftn]: I think we should see if we can remove the "med_present" for this to see if it still works.  I can help test this in the global tests to see if this causes changes/issues if needed.  If we can't test this before this PR is merged we should make an issue to revisit this.

**Review by JessicaMeixner-NOAA — CHANGES_REQUESTED** [review]
@uturuncoglu  thanks again for all of the detailed explanations in the code review.  It really helped my understanding of things. 

I'm going ahead and submitting this review for now, because I think we want to make a few small updates or add issues and wanted to give the feedback I had for now.  I'm still looking more in depth at the data initialize and the time interpolation question for the blending.  We had talked about having a simple field for the background and that could also be one way of just looking at the results to figure that out by say having a constant wind field that just updates to a new constant every hour and see if we get any changes on the sub-hourly scale?

**Comment by JessicaMeixner-NOAA** [comment]
@uturuncoglu I'm looking at the blending routine for the time steps and I'm worried we need to look a little closer into this.  Thoughts on a path forward: 
1. Do you have a run directory with the extra test/debug output turned on that you could share? The info I'm looking for might already be there? 
2. If not, can you point me to the fork/branch of ufs-weather-model with a test that I can run with this blend on one-way coupled from atm to look at things and maybe run my own test? 
We might just need to set up another meeting for this as well to talk through some of the ReadFromFile, but I think 1/2 are going to end up being needed either way.

**Comment by JessicaMeixner-NOAA** [comment]
Also, I think we need to pull together an IC so we can see how this interacts with the WRST switch, I suspect we might have an issue of what is overwriting what. If you can let me know what date you are working with I can create the IC for us to use for testing.

File: model/ftn/wmesmfmd.ftn (hunk 4 above)
Comment (uturuncoglu) [reply model/ftn/wmesmfmd.ftn]: This is working without if statements. So, I removed it and can be used without activating mediator.

File: model/ftn/wmesmfmd.ftn (hunk 6 above)
Comment (uturuncoglu) [reply model/ftn/wmesmfmd.ftn:1930]: Yes, we need to revisit this part once we have a configuration that restarts the model and uses CMEPS for coupling. I think there is no need to take action at this point.

File: model/ftn/wmesmfmd.ftn (hunk 7 above)
Comment (uturuncoglu) [reply model/ftn/wmesmfmd.ftn]: I think this part of code assumes to work on ESMF grid because all the ESMF calls in this subroutine related with grid and also the name of the subroutine is `CreateImpGrid`. There is also export one `CreateExpGrid`. Do we want to make it on all the time? @JessicaMeixner-NOAA what do you think?

Comment (uturuncoglu) [reply model/ftn/wmesmfmd.ftn]: This part of code is working without the check. So, I am removing "med_present" from the part of the code that includes the corner coordinates.

File: model/ftn/wmesmfmd.ftn (hunk 8 above)
Comment (uturuncoglu) [reply model/ftn/wmesmfmd.ftn]: Is this related with the adding corner coordinates? I am asking because there is no such check in the code that adds corner coordinates.

File: model/ftn/wmesmfmd.ftn (hunk 7 above)
Comment (uturuncoglu) [reply model/ftn/wmesmfmd.ftn]: It would be nice to check it with global application.

File: model/ftn/wmesmfmd.ftn [hunk 12]
```diff
@@ -4626,6 +4939,135 @@
...
+      subroutine SetupImpMmsk( mmskField, impField, fillVal, mskCreated, rc )
+!/
+!/                  +-----------------------------------+
+!/                  | WAVEWATCH III           NOAA/NCEP |
+!/                  |           U. Turuncoglu           |
+!/                  |                        FORTRAN 90 |
+!/                  | Last update :         18-May-2021 |
+!/                  +-----------------------------------+
+!/
+!/    DD-MMM-YYYY : Origination.                        ( version 7.XX )
```
JessicaMeixner-NOAA: add a date to fille in DD-MM-YYYY

model/ftn/wmesmfmd.ftn
JessicaMeixner-NOAA: I think this could be a fairly major issue here.  If it's not addressed now, there needs to be an issue documenting that this is happening and that we need to look into it further.

model/ftn/wmesmfmd.ftn
uturuncoglu: @JessicaMeixner-NOAA if you look at following line of code https://github.com/NOAA-EMC/WW3/blob/3740ba99caaa2822b5098bf32faeaf199a710a51/model/ftn/w3fldsmd.ftn#L1118 the time interpolation is not enabled intentionally with the coupling (`FLINTERP = .FALSE.`). I did not test it but it seems like that. So, if we want to force it and if it is really related with the time interpolation then we need to set `FLINTERP = .TRUE.`. Let me know what do you think?

model/ftn/wmesmfmd.ftn
uturuncoglu: As @BinLiu-NOAA is suggested I tried to use `if (abs(dptr(i,j)-fillVal) < 1.0e-8) then` but this is failing and I think that it is not creating merge mask as expected. I'll also plaining to test equal. @danrosen25 Let me know if you have any other suggestion.

uturuncoglu: BTW, the type of the dptr and fill Value is `ESMF_KIND_RX`which is R4 in WW3 cap by default. In CMEPS the type is R8. So, this could be a source of the problem and explains why @BinLiu-NOAA suggestion is not working. Maybe it is better to keep as its.

uturuncoglu: I tested with equal and it is failing too. So, at this point it is better to keep as it is. We could try to build model with R8 but as I know it is not easy under UFS model and R8 needs to be default for CMEPS coupling of course if it works.

model/ftn/wmesmfmd.ftn
uturuncoglu: I am testing to set `flagsc = .true.` in w3fldg call. This will activate the time interpolation. I'll update you with the results.

uturuncoglu: I think that setting `flagsc = .false.` is fine and this will activate the time interpolation if the field is 'CUR', 'WND' (our case) 'WNS', 'TAU' or 'RHO'.  So, this is current behavior in the model but if you look at the following figure (red is from the point in the active coupling region and white is from data - outside of the active coupling region),

Screen Shot 2021-06-24 at 3 31 05 PM

and it seem that there is no time interpolation (white one) and it is keeping constant in 3-hourly interval which I think it is the temporal resolution of the wind forcing. Anyway, I am plaining to put some print statement to see the value of `FLINTERP`. Another possibility could be having same data in current and next time and then the temohtal interpolation will produce constant value for 3-hours. I'll also check that one too.

model/ftn/wmesmfmd.ftn
uturuncoglu: do you mean date for Origination?

uturuncoglu:
@JessicaMeixner-NOAA I run the model couple of hours (atm-wav configuration with one way interaction) you could find the results in the following directory on Orion `/work/noaa/gmtb/tufuk/RUN/AtmWavOneDir/save`. Please let me know if you could not reach it.

uturuncoglu:
I think that I have correct wind IC for wave. Once we realized that we are forcing the wave model with the data that belongs to different hurricane other than Dorian. @BinLiu-NOAA created new forcing for me for Dorian case using HWRF workflow. So, if you think that we have still issue with the wave IC, just let me know.

model/ftn/wmesmfmd.ftn
uturuncoglu: I print out `FLCOUPL` and `FLINTERP`. The `FLCOUPL` is set to False as expected and consistent with the `flagsc = .false.`. The `FLINTERP` seems to set as True. So, I was expecting to have time interpolation for wind at this point but previous plot does not support it. Maybe I am missing something because it is hard to follow `W3FLDG` call. @JessicaMeixner-NOAA do you have any idea about it.

uturuncoglu: I also checked the w*0 and w*n fields read from the forcing file and their min and max values. It seems that they are fine. So, I think that read call gets the current and next time steps correctly from file and maybe there is a issue in time interpolation. 

```
20210624 172141.669 INFO             PET1340    time0 =     20190829           0
20210624 172141.669 INFO             PET1340    timen =     20190829         600
20210624 172141.669 INFO             PET1340    tw0 =     20190829           0
20210624 172141.669 INFO             PET1340    twn =     20190829       30000
20210624 172141.670 INFO             PET1340    wx0 min, max =   -19.55906       14.35927
20210624 172141.670 INFO             PET1340    wy0 min, max =   -16.34938       19.25990
20210624 172141.671 INFO             PET1340    wxn min, max =   -22.29049       13.06400
20210624 172141.671 INFO             PET1340    wyn min, max =   -15.75515       20.08858
```

Also time interval considered (time0 and timen) and field times (tw0 and twn) are fine too. I am not sure about the time interpolation logic and it seems that we need to look at it more detailed way.

JessicaMeixner-NOAA:
@uturuncoglu thanks for being patient with me as I understand all of this better.  I was able to access your directory and will look at the results more closely today which I think will help me see what you have commented here.

model/ftn/wmesmfmd.ftn
JessicaMeixner-NOAA: Thanks for this explanation and the rundir.  I'll look at this today to make more sense of it.  Usually for WW3 when we read in from the file, we only read in a new file time for twN once we have gotten to that, and then we set tw0=twN and same for the wind values and then just get the new winds for the twn time.  The fact that this is called every time regardless and that it's set to zero each time (and not just for the first call) is what made me look twice.   Again, I'll look at the output more closely today so I can try to understand your explanations better.  Thanks again for sharing the output!

uturuncoglu:
@JessicaMeixner-NOAA Sure. Let me know if you find anything that needs to be fixed.

model/ftn/wmesmfmd.ftn
JessicaMeixner-NOAA: Apologies, I did not look at this last week.    I've looked at it now and with your explanations and looking at the run directory and plotting I see that it's working.  What through me off is that typically when we are reading from a file you do not set all of the time and variables to zero each time step and we use prior information and tw01=twN when current time gets to twN and then you don't re-read those fields.  I think this might be something we could revisit if efficiency updates are need or if we decide we want to try to interpolate the wind between the snapshots in this region.

JessicaMeixner-NOAA (commented):
I made one more comment noting an area where we can improve on efficiency if needed in the future.  Since both WW3 and ufs-weather-model have updated since the last time the ufs-weather-model regression tests were run using this branch, we should probably do that one more time before merging.  Otherwise, I think this is good to go.  Thanks Ufuk and sorry for the delay in my review.

model/ftn/wmesmfmd.ftn
uturuncoglu: @JessicaMeixner-NOAA No worries. I know you are too busy with the PRs etc. Yes, we could still revisit this part of code. I could not find detailed information about the time interpolation in the read routine but if somebody has more information about that we could also enable temporal interpolation. Another option could be creating wind forcing file by increasing its temporal frequency by performing temporal interpolation outside. So, I know that the default data is 3-hourly but if we could modify the script that creates the wind input and add an option to it to perform temporal interpolation to create i.e. 1 -hourly forcing file then we don't need to change anything internally in the cap. We could just use already interpolated forcing.

model/ftn/wmesmfmd.ftn
uturuncoglu: @danrosen25 if you don't have any objection about it, I'll resolve this.

model/ftn/wmesmfmd.ftn
JessicaMeixner-NOAA: Sounds good! Always nice to have a few options so we can figure out what is most efficient or most flexible, etc.

JessicaMeixner-NOAA:
@uturuncoglu I ran the ufs-weather-model regtests w/your branch (after merging in WW3 develop) and the regression tests did not pass.  I didn't get a chance to figure out why, but wanted to let you and @aliabdolali know so we can look into that before merging this PR.

uturuncoglu:
@JessicaMeixner-NOAA as remember correctly, it was passing before right? I also remember that I run full RTs and it was passing. We have following possibilities:

- @BinLiu-NOAA synced the branch and it could be an issue over there. 
- It could be related to the recent changes that I did after your PR comments. We could still go back and test the old version to find the commit that causes issue. 

Let me look at those and I'll update you about it.

uturuncoglu:
@JessicaMeixner-NOAA which test can I test

uturuncoglu:
Sorry, I closed and reopen the PR. It is my fault.

uturuncoglu:
@JessicaMeixner-NOAA I run the `cpld_control_wave` and it fails with following trace that goes to WW3 cap. 

```
/work/noaa/gmtb/tufuk/HAFS_wav/sorc/hafs_forecast.fd/WW3/model/tmp/wmesmfmd.F90:1384
/work/noaa/gmtb/tufuk/HAFS_wav/sorc/hafs_forecast.fd/WW3/model/tmp/wmesmfmd.F90:3109
```

This is the line of code that perform halo update,

```
!
! 3.b Store import field halo routeHandle
!
      call ESMF_FieldHaloStore( impMask, routeHandle=impHaloRH, rc=rc )
      if (ESMF_LogFoundError(rc, PASSTHRU)) return
```

So, it does not makes sense to me. Anyway, I'll try to roll back my recent changes to see what happens. The only think that is in my mind is the commit that removes control to add corner coordinates to import and export grid. Anyway, i'll update you about it.

uturuncoglu:
I go back just before commit that removes the check for adding corner coordinates and model seems working and this particular test passes. So, it seems that https://github.com/hafs-community/WW3/commit/f4929723c63a46aed729239f6b7faa253a4c0c86 commit has the issue if we run it with connectors. At this point I am not sure why it is failing. Maybe halo region is not defined correctly in corner coordinates or maybe it is an issue in ESMF side. I'll check and let you know.

uturuncoglu:
@JessicaMeixner-NOAA I think I solved the issue. Please see following commit https://github.com/NOAA-EMC/WW3/pull/396/commits/1f1ed0714a4575bba3bec8b58d3afb053f2b9890. I also run `cpld_control_wave` and it passes now. I could also run other wave configurations but I am not expecting any issue with others. I'll also check the HAFS wave configurations too and let you know. BTW, the debug section in `CreateImpGrid` and `CreateExpGrid` just creates tmp field in center stagger to write out grid coordinates. It might be useful to do the same for corner stagger. We could create another subroutine similar to [this call used in CMEPS](https://github.com/ESCOMP/CMEPS/blob/22dc297fa3f77cf1a232618f5428c89b460032ca/mediator/med.F90#L2648) to write all the grid coordinates, mask etc. to a single file. This seems more elegant way rather than writing them into individual files. Let me know what do you think?

model/ftn/wmesmfmd.ftn
uturuncoglu: @JessicaMeixner-NOAA new issue - https://github.com/NOAA-EMC/WW3/issues/428 - is created to track it.

uturuncoglu:
@JessicaMeixner-NOAA I run all wave related regression tests (cpld_bmark_wave, cpld_bmarkfrac_wave, cpld_bmarkfrac_wave_v16 and cpld_control_wave) and all of them were passed on Orion. I also tested fully coupled FV3+WW3 regional configuration (two-way interaction) under HAFS applications and it is also working without any issue.

JessicaMeixner-NOAA:
@uturuncoglu thanks for tracking down and solving the issue.  I'll retest on hera today to confirm your results on orion.

JessicaMeixner-NOAA:
After hera maintenance, I ran the full ufs-weather-model regression tests with hera.intel with this branch of WW3 + merging in the develop branch and one test failed: 
Test control_atm_aerosols 093 failed in run_test failed
@uturuncoglu you you just ran the wave tests on orion, correct? I feel fairly confident that atm->aerosols was not effected by this change in WW3.  I'm going to re-run just that test to see if it was a one-off error.

JessicaMeixner-NOAA:
re-running the failed control_atm_aerosols test, it passed this time w/hera.intel

JessicaMeixner-NOAA (approved):
A group review of this code was completed and outstanding items were either addressed or issues for follow up were made. 

The ufs-weather-model (using commit hash b64e648c34bd177a761195fe0bf17da6be44fc09) + this branch with the WW3 develop merged in passed all regression tests for me on hera.intel  

Thanks @uturuncoglu for all of the detailed explanations and these updates!!!!

uturuncoglu:
@JessicaMeixner-NOAA that is great news. I think that PR is in better shape now with your comments and suggestions.

aliabdolali:
The regtest passed successfully on hera using intel,impi, with expected known non-b4b cases:

```
mww3_test_03/./work_PR2_UQ_MPI_d2                     (9 files differ)
mww3_test_03/./work_PR2_UNO_MPI_d2                     (8 files differ)
mww3_test_03/./work_PR3_UQ_MPI_d2_c                     (8 files differ)
mww3_test_03/./work_PR3_UQ_MPI_d2                     (8 files differ)
mww3_test_03/./work_PR3_UNO_MPI_d2_c                     (8 files differ)
mww3_test_03/./work_PR1_MPI_d2                     (8 files differ)
mww3_test_03/./work_PR3_UNO_MPI_d2                     (6 files differ)
mww3_test_07/./work_PR3_UQ                     (3 files differ)
ww3_tp2.10/./work_MPI_OMPH                     (6 files differ)
ww3_tp2.15/./work_5km                     (1 files differ)
ww3_tp2.15/./work_MPI_5km                     (1 files differ)
```

[matrixCompFull.txt](https://github.com/NOAA-EMC/WW3/files/6808581/matrixCompFull.txt)
[matrixCompSummary.txt](https://github.com/NOAA-EMC/WW3/files/6808582/matrixCompSummary.txt)
[matrixDiff.txt](https://github.com/NOAA-EMC/WW3/files/6808583/matrixDiff.txt)

aliabdolali:
@ danrosen25 could you provide your review, so we can move on and merge this PR.

uturuncoglu:
@danrosen25 do you have anything that needs to be addressed?

danrosen25 (approved):
I haven't been able to run regression tests but I'm holding up the completion of this PR. I am not keen on the use of med_present as this creates an end-user configuration that needs documentation for a package of features that seem to be independent on med_present=t/f. There are still a few things that will need to be addressed such as data dependencies during initialization

JessicaMeixner-NOAA:
@danrosen25 @uturuncoglu are we in such a rush that we need to merge things we aren't okay with?  At minimum can we make an issue to address the "if med_present" issue?  What are the alternatives for getting rid of this if- mediator statements?    

We currently have an issue for the initialization: 
https://github.com/NOAA-EMC/WW3/issues/428 and documentation https://github.com/NOAA-EMC/WW3/issues/399 (@uturuncoglu should we go ahead an assign those to you?) 

With not running the regression tests, was this a time issue or a the regression tests aren't working issue?

uturuncoglu:
@JessicaMeixner-NOAA I think that we already remove most of `med_present` such as from adding corner coordinates etc. I think removing all of them will take time since I have no too much time under HAFS project. I need to wrap-up everything including CDEPS, WW3 work until end of this month. Of curse you could assign those issues to me but I am not sure when I will find time to work on those because I am supposed to work on UFS-xgrid implementation at the beginning of August. I am not sure about you questions related to the RTs. As I know  the wave related RTs are passing without any issue.

uturuncoglu:
@JessicaMeixner-NOAA currently `med_present` only used for following,

- adding scalar field to export state. this is CMEPS specific and required to have 2d fields in mediator history and restart.
- not checking field time stamps in Data initialization, which is related with the https://github.com/NOAA-EMC/WW3/issues/428 issue.

So, we don't have too much related with `med_present` at this point.

danrosen25:
@JessicaMeixner-NOAA
There will be a progress report next week for hurricane supplemental progress and if we can close this before the report that would be better.

JessicaMeixner-NOAA:
Sounds good.. Note that there needs to be a separate ufs-weather-model PR for this or you can work with @aliabdolali as he has an open ufs-weather-model PR that involves updating WW3 to the top of develop that is not yet merged

Issue #360, opened 2021-04-20
Transition from switch pre-processing to CPP ifdefs


JessicaMeixner-NOAA:
Instead of using the current paradigm of choosing switches, and then processing the *.ftn code to *.F90 code, we will convert all code to *.F90 and use the switches to define CPP if defs.  

This work is planned for June 2021.  To prepare for this, we recommend making pull requests with updates as soon as possible. In addition, removal of switches that are no longer needed have been proposed and will be cleaned out first.

PR #405, opened 2021-06-10
PREVIEW: This is a first pass at change from switches to CPP 

JessicaMeixner-NOAA:
# Pull Request Summary

This is a first pass of going  from switches to CPP.  There is not a make system for this yet.  The new src folder lives alongside the ftn directory.  We'd probably want to do a "git mv ftn/*.ftn src/*.F90" in attempts to keep some git history, but this is just to get some feedback as we go along.  See issue https://github.com/NOAA-EMC/WW3/issues/360

* aux/switch2cpp.awk is from @ukmo-ccbunney that was used to convert the code from *ftn to *F90 
* ftn2src.sh is a wrapper to convert all ftn files

I'm making the PR in attempts to get comments and feedback as early as possible.  In terms of timing the first step will be to finish the clean up (see: https://github.com/NOAA-EMC/WW3/projects/1) and then this (complete with gnu build). 

This PR is not intended to be merged.  Likely another PR will be what we end up merging but this is here to get the conversation started. 

*** DO NOT MERGE ***


JessicaMeixner-NOAA:
I think this is an example that shows the renaming: https://github.com/NOAA-EMC/WW3/compare/develop...JessicaMeixner-NOAA:feature/ftn2src?expand=1

model/src/ww3_ounf.F90
```diff
@@ -0,0 +1,4112 @@
...
+      IYN = NYO
+     
+      ! Also store NXO and NYO in __local__ RTDNX and RTDNY variables.
+      ! This avoids compilation errors when the RTD switch is enabled
+      ! but the SMC switch is not. TODO: Remove this when C-preprocessor
+      ! is used in preference to switches.
+      RTDNX = NXO
+      RTDNY = NYO
+
+!/RTD      ! SMC type 3/4 outputs are currently on standard pole grid only
```
ukmo-ccbunney: @JessicaMeixner-NOAA One thing I did not factor into my awk conversion script is the conversion of lines that contain double switches, like this one (the second `!/RTD` switch is still present).

E.g. lines that look something lie this:
`!/SMC!/RTD   ! This is a double switch line`

In these cases, to avoid making the awk script overly complicated, I found it easier to simply rerun the conversion script until it reports no more converted switches.

ukmo-ccbunney: You could add an extra loop in your control script to test for the existence of any switches in the input file, something like this (unteted!) around line 18 of `ftn2src.sh`:

```sh
.
.
       echo "convert $file.ftn to a .F90"   

       cp -f $ftndir/${file}.ftn $ftndir/${file}.tmp

       # Keep processing file until no switches remain:
       while $(egrep "^\!/[0-9A-Za-z]+" ${file}.tmp); do
          gawk -f $thisdir/switch2cpp.awk < $ftndir/${file}.ftn > $srcdir/${file}.F90
          cp -f $ftndir/${file}.F90 $ftndir/${file}.tmp
       done

       rm -f  $ftndir/${file}.tmp
.
.
```

JessicaMeixner-NOAA: @ukmo-ccbunney  thanks for catching the double switch issue! I'll add a second loop to take care of this.  Is there any issue in just running the switch2cpp.awk on every file twice?

ukmo-ccbunney: >  Is there any issue in just running the switch2cpp.awk on every file twice?

I don't think so - it should simply report "**Converted 0 switches...**" for most files that don't have double switches.
Now - I don't think there any triple switches...but maybe we should check :)

JessicaMeixner-NOAA: Okay, I've updated for processing 3 times to be extra cautious.  I also updated this PR to show the changes in the ftn directory because I think that's easier to see changes then too.

JessicaMeixner-NOAA:
@ukmo-ccbunney I know one reason you said that you added the w3 in front of all of the switches that there were conflicts with compiler switches.  Do you by chance remember what those were?  I'm wondering because at the tops of routines and other places we have things like "switches used" and they don't have the W3_<SWITCH> there.  Does that get confusing as we move forward for new users?  I'm wondering if it would be worth while to switch the names of switches that have conflicts? 

A positive of having the W3_switch is that it's clear, particularly in coupled modeling environments, that this is a WW3 switch which is a nice thing.  

Thoughts?

ukmo-ccbunney:
> 
> 
> @ukmo-ccbunney I know one reason you said that you added the w3 in front of all of the switches that there were conflicts with compiler switches. Do you by chance remember what those were? I'm wondering because at the tops of routines and other places we have things like "switches used" and they don't have the W3_ there. Does that get confusing as we move forward for new users? I'm wondering if it would be worth while to switch the names of switches that have conflicts?


Hmm.. not sure I remember which switches were conflicting, or whether I did is a pre-emptively to avoid this happening in the future.

As for the "switches used" comments, I am sure I can come up with some awk/regex scripting that could catch these comment lines and update the switches (from a known list of existing switches). I'll look into it when I have a chance.

JessicaMeixner-NOAA:
@ukmo-ccbunney thanks!   I was actually wondering if having the ones that do correspond to compiler commands, such as some of the OMP ones, would actually be a good thing to keep too?

ukmo-ccbunney:
> 
> 
> @ukmo-ccbunney thanks! I was actually wondering if having the ones that do correspond to compiler commands, such as some of the OMP ones, would actually be a good thing to keep too?

Maybe...although I prefer the idea of everything having the same prefix.
I don't have a strong opinion on this though - as long as it does not make for any "special case" handling of those #defines in any associated scripting.

JessicaMeixner-NOAA:
I see the benefits of everything having the same pre-fix.  The things that gave me pause were: 
* comments in code (or places where there was a commented out switch... I can't point you to an example now but I've seen a few places in code where we have ! !/SWITCH, maybe these just eventually slowly get cleaned out)
* do we then make our switch files add those W3_ prefixes or leave them as is?  
* How do we make things as consistent as possible?

JessicaMeixner-NOAA:
@ukmo-ccbunney the script "model/bin/all_switches" seems to still work to give a list of all switches currently used: 

    ARC
    B4B BS0 BS1 BSX BT0 BT1 BT2 BT3 BT4 BT8 BT9 BTX
    COU CRT0 CRT1 CRT2 CRX0 CRX1 CRX2
    DB0 DB1 DBX DEBUG DEBUGCOH DEBUGDB1 DEBUGDCXDX DEBUGFLS DEBUGFREQSHIFT DEBUGGRID DEBUGINIT DEBUGIO DEBUGIOBC DEBUGIOBP DEBUGIOGR DEBUGMPI DEBUGNETCDF DEBUGOUTPUT DEBUGREFRACTION DEBUGRUN DEBUGSETIOBP DEBUGSETUGIOBP DEBUGSOLVER DEBUGSOLVERCOH DEBUGSRC DEBUGSTP DEBUGW3ULEV
    del
    DIST DSS0 DUM
    EXPORTWWM
    F F90 FLD1 FLD2 FLX0 FLX1 FLX2 FLX3 FLX4 FLXX FT
    IC0 IC1 IC2 IC3 IC4 IC5 IG1 IS0 IS1 IS2
    LN0 LN1 LNX
    MEMCHECK MGG MGP MGW MLIM MPI MPIBDI MPIT MPRF
    NC4 NCEP2 NCO NETCDF_QAD NL0 NL1 NL2 NL3 NL4 NLS NLX NNT NOGRB
    O0 O1 O10 O11 O12 O13 O14 O15 O16 O2 O2a O2b O2c O3 O4 O5 O6 O7 O7a O7b O8 O9 OASACM OASASCM OASICM OASIS OASOCM OMP0 OMPG OMPH
    PDLIB PR0 PR1 PR2 PR3 PRX
    REF1 REFRX REFT RTD RWND
    S SBS SCRIP SCRIPNC SEC1 SEED SETUP SHRD SMC SMCT ST0 ST1 ST2 ST3 ST4 ST6 STAB2 STAB3 STX
    T T0 T1 T1a T2 T3 T38 T4 T5 T6 T7 T8 T9 TDYN TIDE TIDET TIMINGS TR0 TR1 TRKNC TRX
    UNO UOST UQ
    WCOR WNT0 WNT1 WNT2 WNX0 WNX1 WNX2 WRST
    XW0 XW1 XX0 XXX

JessicaMeixner-NOAA:
When updating the build system, one key aspect is to  change https://github.com/NOAA-EMC/WW3/blob/develop/model/bin/ad3#L230 
from  a call to w3adc to process the switches 

`$path_b/w3adc < w3adc.inp.$name > w3adc.out.$name 2> w3adc.err.$name
`
to a call that processes the cpp flags so something like: 

`gfortran -DW3_switch1 -DW3_switch2 -E $path_i/$name.F90 -o $name.F90 
`
however, when trying to do this I'm getting errors because of things like this: 

`#endif // W3_T
`
I can change in the awk script to not have the `// W3_T` but we also end up getting `#endif !` lines, which also cause error.  I might have the preprocessesing the cpp flags command incorrect to get the temporary file, or I might have incorrectly modified the awk script or both.   Any insight from others? 
@ukmo-ccbunney @aliabdolali ??

ukmo-ccbunney:
Ah - that's probably my fault. I think that the comments should probably be standard C style, rather than newer C++/C90
I.e.

```C
#ifdef WW_SW1
! This comment will be included
#endif /* WW_SW1 */
```

rather than:

```C
#ifdef WW_SW1
! This comment will be included
#endif // WW_SW1
```

For me, the ifort compiler accpets both without complaint. However, gfortran produces a warning for the second form (but not an error).

ukmo-ccbunney:
Also, I wonder if we should be invoking "cpp" directly here, rather than doing it via compiler flags (which will be different per compiler)?

I.e. `cpp -DW3_switch1 -DW3_switch2 $path_i/$name.F90 > $name.F90 `

Just a thought...?


**EDIT: Although - after saying that I notice that IFORT uses it's own "fortran pre-processor" (FPP) to pre-process, so perhaps it is better to let the compiler choose it's own preferred pre-processor.**

JessicaMeixner-NOAA:
@ukmo-ccbunney I go back and forth with this:


> Also, I wonder if we should be invoking "cpp" directly here, rather than doing it via compiler flags (which will be different per compiler)?
> 
> I.e. `cpp -DW3_switch1 -DW3_switch2 $path_i/$name.F90 > $name.F90 `
> 
> Just a thought...?
> 
> **EDIT: Although - after saying that I notice that IFORT uses it's own "fortran pre-processor" (FPP) to pre-process, so perhaps it is better to let the compiler choose it's own preferred pre-processor.**

We need something for the w3init routine (although we could have a different work around for the putting in of the switch files) and then we need something that does the cpp -> processed file for the creation of the dependencies of the subroutines, but then we could still pass the flags into the comp and let the compiler do it's own preferred pre-processing?

JessicaMeixner-NOAA:
Closing this PR... Please see: https://github.com/NOAA-EMC/WW3/pull/476

PR #475, opened 2021-09-15
Use CPP instead of Switches (no src, just build) 

JessicaMeixner-NOAA:
# Pull Request Summary
Update from the WW3 switch pre-processor to CPP.  Build info only.  This is a draft PR and the other PR should be used instead, this is just for an example of what the build itself will be and is a good PR to merge in if you need to convert ftn files to src for work on existing code/branches. 

## Description
This PR provides a utility in tools/ that converts the WW3 switches into CPP ifdefs.  The build is also updated for this change. This PR does not include the change from ftn--> src files.  (Another PR with this change will be opened, but this will help isolate the changes that are just build related). 

Co-Author: @ukmo-ccbunney 

### Issue(s) addressed
* fixes Issue #360 

### Check list  
* Is your feature branch up to date with the authoritative repository (NOAA/develop)? yes (+ PR#454 is included)
* Make sure you have checked the [checklist for a developer submitting to develop](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-a-developer-submitting-to-develop) and [updating version number](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-updating-version-number) 

Should we update the version number for this? It's a big change? Currently the version number is not updated. 

Regtests are not done on this PR as the ftn -> src conversion is not included here.


ukmo-ccbunney:
@JessicaMeixner-NOAA I've tried running this and it works well for my **GNU** compiler, but fails for the **Cray Fortran** compiler.
The reason being that the compile flag for "perform source pre-processing" is `-eP`, not `-E`.
Also, annoyingly, if you specify the `-eP` flag, you can't specify the `-o` flag; the output always goes to `file.i`

Essentially, I had to modify line 230 of `ad3` to this for the Cray Fortran compiler:

```
    #$comp_fc $CPPFLAGS -E $path_i/$name.F90 -o $name.F90
    $comp_fc $CPPFLAGS -eP $path_i/$name.F90 #-o $name.F90
    mv $name.i $name.F90
```

Perhaps we need to define some kind of compiler specific CPP _hook function_ that ad3 can call for each file? 
The function gets an input filename and should do whatever needs to be done to produce a pre-processed file called `file.F90`.
Perhaps it could be set up as part of the `w3_config` step using the cmplr.env file?

JessicaMeixner-NOAA:
@ukmo-ccbunney thanks for testing and let me know! 

We almost exclusively use gfortran for the pre-processing even if we use another compiler for actually compiling the code, which is an odd combination I admit.   I like the idea of having something in the cmplr.env instead, which removes the need to define comp_fc in the first place. 

The only reason we have to have the $name.F90 file is because of the w3init file (see discussion topic here: https://github.com/NOAA-EMC/WW3/discussions/474); If we didn't have that --  then we could remove the need for ad3 at all and just move the CPP to part of the compilation.  But I've had no feedback on that discussion item and don't want to unilaterally make that decision to just get rid of it.

JessicaMeixner-NOAA:
@ukmo-ccbunney I added a if-block in ad3, I don't think this is the cleanest way to do it, but I haven't yet figured out how to add this into the cmplr.env either yet.  I think we could remove the w3init issue this could be moved to cmplr.env and cleaned up/simplified further.  Let me know what you think.

ukmo-ccbunney:
> @ukmo-ccbunney I added a if-block in ad3, I don't think this is the cleanest way to do it, but I haven't yet figured out how to add this into the cmplr.env either yet. I think we could remove the w3init issue this could be moved to cmplr.env and cleaned up/simplified further. Let me know what you think.

HI @JessicaMeixner-NOAA - I don't think that will work as you intended unfortunately. In the Cray Programming Environment, the "ftn" command is just a wrapper around whatever compiler modules you have loaded (Cray, GNU, Intel, etc); it passes most of the switches it receives to the underlying compiler. I call "ftn" regardless of the underlying compiler that is being used, therefore we can't really test on `comp_fc == "ftn"`

Following up on the conversation in #474, I would personally prefer if we could find an alternative solution to populating the switches in w3initmd so we can do away with ad3. Can we not do this with a CPP macro instead?

I will have a look later today to see if I can find a CPP solution.

ukmo-ccbunney:
@JessicaMeixner-NOAA I have an edit to `ad3` and `w3initmd.f90` that enables us to set the switch string with a -D flag, if that is useful.

I changed lines from 225 in ad3 to this:
```
    #Generate list of CPP Flags based on switches:
    CPPFLAGS=()
    for sw in $sw_str
    do
       CPPFLAGS+=("-DW3_$sw")
    done 
    
    CPPFLAGS+=("-D__WW3_SWITCHES__='$sw_str'")

    $comp_fc "${CPPFLAGS[@]}" -E $path_i/$name.F90 -o $name.F90
```

and added this to w3initmd.F90 (line 116/117):
```
      CHARACTER(LEN=512), PARAMETER  :: SWITCHES  = &
          __WW3_SWITCHES__
```

NB: the use of a bash array for CPPFLAGS was necessary to get around an issue I was having with the quotes around the space separated `__WW3_SWITCHES__` variable

ukmo-ccbunney:
@JessicaMeixner-NOAA Above code pushed to branch:  https://github.com/ukmo-waves/WW3/tree/feature/switch2cpp_w3initmd
Commit: https://github.com/ukmo-waves/WW3/commit/fe2d60e11abc85edce416ce8a7a14aa4f603d0a0

JessicaMeixner-NOAA:
@ukmo-ccbunney I have added your update to ad3 and it worked for me as well and have started looking at if we could just get rid of ad3 completely and I think the answer is maybe, but it's not going to be a simple, because ad3 handles copying the *.o and *.mod files, which to keep things as is we'd need to move to comp which I don't know if we want to do or not?   

Even if we don't get rid of ad3, we could export the switch definitions and then add the CPP flags into comp/cmpl.env which is probably what we want to do, regardless of if we get rid of ad3 or not.  

To simplify our build further I think the hold up is the fact that we compile some programs serially versus others with MPI and/or OMP and we want to do that without recompiling things all the time for the regression tests and therefore have the obj_MPI mod_MPI etc folders.  If we did not have to worry about that, getting rid of ad3 is simpler. 

So I guess my question for everyone is:  Do I try to continue to get rid of ad3 or do we just move the CPP processing to be done within comp and then worry about further simplifying our build by looking at compiling the pre/post processing jobs with MPI/OMP (even if we aren't necessarily running/taking advantage of the OMP/MPI) and also perhaps moving to a CMAKE build?   I  am leaning towards not trying to get rid of ad3 for now, but I'm open to other paths forward.

ukmo-ccbunney:
@JessicaMeixner-NOAA If we consider this first tranche of work to be focused on using CPP instead of WW3 style switches, then I think that this is close the being achieved satisfactorily. Perhaps we should consider that the removal of ad3 be tackled separately at a later date, considering the complexities of serial/MPI/OMP builds you mention above? Especially so if we consider moving to CMAKE which I assume would be quite  a different build system again?

JessicaMeixner-NOAA:
@ukmo-ccbunney actually until we over-haul the build even further we cannot get rid of ad3 because it's used in make_makefile when the dependencies are created.  So I think we'll have to leave things as is for now, although I did manage to do a little more clean-up.  Does this work for you now, or is it still not working for you? Do you need other changes to work with the cray compilers?

JessicaMeixner-NOAA:
@ukmo-ccbunney -- if the current ad3 does not work for the cray compilers, what we could do is remove the dependency on   comp_fc=$WWATCH3_F90
and have a templated file just like we do for comp/link and ad3 -> ad3.tmpl and then line: 
$comp_fc "${CPPFLAGS[@]}" -E $path_i/$name.F90 -o $name.F90
could be replaced by a variable that then gets replaced by the correct command based on which compiler you choose in cmpl.env when you run w3_setup

JessicaMeixner-NOAA:
I tried to use a template for ad3 in this branch here: https://github.com/JessicaMeixner-NOAA/WW3/tree/feature/ad3template but it's not working for me correctly yet, but this is more/less the idea

JessicaMeixner-NOAA:
Closing this PR for now.

PR #641, opened 2022-03-20
Add mesh cap 

### LINKED ISSUE #636: Add mesh-based nuopc cap to the dev/ufs-weather-model branch
**Is your feature request related to a problem? Please describe.**

The existing NUOPC cap is not restart reproducible in for the S2SW application of the UFS Weather model. The existing cap also cannot be run in debug mode for the S2SW application.  

**Describe the solution you'd like**
A new mesh based NUOPC cap has been developed which will be used by the UWM S2SW and HAFSW applications, both of which use the dev/ufs-weather-model branch. The mesh cap is restart reproducible for UWM and can be run in debug mode.  

For the S2SW app, the mesh cap will replace the current NUOPC connectors and coupling will go through the CMEPS mediator. This allows the wave model to be placed in the slow coupling loop if desired. The HAFSW application currently uses the CMEPS mediator, but will now run on a mesh inside the NUOPC cap in place of a grid.

The current ATMW applications will continue using connectors.

**Describe alternatives you've considered**

**Additional context**
- https://github.com/ufs-community/ufs-weather-model/issues/739
- https://github.com/ufs-community/ufs-weather-model/issues/229
- https://github.com/NOAA-EMC/WW3/issues/552

DeniseWorthen:
# Pull Request Summary

Enables an alternate cap for use by UWM apps S2SW and HAFSW

## Description

The alternate source code utilizes the switch file ``switch_meshcap`` . A new switch option (``UWM``) is added to control features required by UWM.

The cap has a run-time config variable which determines whether the wave model is run in shel vs multi mode in UWM. Only a single grid has been verified when run in multi.

The new source files are:
- *wav_comp_nuopc*: This is the main cap code. It advertises and realizes connected fields on a provided ESMF mesh file. 
- *wav_import_export*: This fills import and export state fields.
- *wav_shel_inp*: Enables reading ``ww3_shel.inp`` file when run as shel.
- *wav_shr_mod*: A set of utility routines to reduce code duplication.
- *w3iogoncdmd.F90*: Writes gridded output fields as netcdf. Only a sub-set of all possible gridded output is currently available. When netcdf output is desired, an additional option needs to be added to the switch file (``UWMNCOUT``). This feature was originally coded to allow control via configuration variables but the code changes were backed out in order to resolve issues seen on the CESM side. A refactoring of the ``w3iogoncdmd.F90`` to allow all output variables is currently under development.
- *wav_kind_mod*

Code specific to CESM is added within ``ifdef CESMCOUPLED``. Code specific to UWM is added within ``ifdef W3_UWM``.

Please also include the following information: 
* Add any suggestions for a reviewer 
* Mention any labels that should be added:  _bug_, _documentation_, _enhancement_, _new feature_
* Are answer changes expected from this PR? Please describe the changes and the reason why in addition to which of the following labels would apply: _mod_def change_, _out_grd change_, _out_pnt change_, _restart file change_, _Regression test_ 
-->

### Issue(s) addressed
- fixes #636

### Commit Message

Adds source code to implement a mesh-based NUOPC cap in the dev/ufs-weather-model branch.

co-author: @mvertens

### Check list  

### Testing

* Please indicate the expected changes in the regression test output ([Note the known list of non-identical tests](https://github.com/NOAA-EMC/WW3/wiki/How-to-use-matrix.comp-to-compare-regtests-with-master#4-look-at-results)).


model/src/wav_comp_nuopc.F90
```diff
@@ -0,0 +1,1416 @@
+module wav_comp_nuopc
+
+  !/ ------------------------------------------------------------------- /
+  !/
+  !/                  +-----------------------------------+
+  !/                  | WAVEWATCH III           NOAA/NCEP |
+  !/                  |           H. L. Tolman            |
```
JessicaMeixner-NOAA: Thanks for adding the WW3 header here.  The spot where you have "H.L. Tolman" should be the author(s) of this subroutine.  As I don't think Hendrik wrote this, so I think this needs to be updated.  In addition, I think the "last update:" date should be updated to today's or a recent date. 

A similar header should be included with every module/subroutine that is added.

model/src/w3initmd.F90
```diff
@@ -1236,6 +1246,9 @@ SUBROUTINE W3INIT ( IMOD, IsMulti, FEXT, MDS, MTRACE, ODAT      &
 !
       FLOUT(2) = NPT .GT. 0
 !
+#ifdef CESMCOUPLED
```
JessicaMeixner-NOAA: Delete commented out code?

model/src/w3updtmd.F90
```diff
@@ -1803,6 +1807,9 @@ SUBROUTINE W3UICE ( VA )
                                   ICEI(IX,IY), 'ICE (NEW)'
 #endif
                 VA(:,JSEA) = 0.
+#if defined(W3_UWM) || defined(CESMCOUPLED)
+               charn(jsea) = aalpha
```
JessicaMeixner-NOAA, line 1811: Curious why this and the below update was added?  I'm wondering if this should really be behind switches or just actual additions to the code?

JessicaMeixner-NOAA (commented):
@DeniseWorthen thanks for making this PR! There's a lot of awesome updates here. I've made a few comments and am continuing to look at and review the code, but wanted to ask a few questions now. 
* I know this is a WIP, but I'm curious when we might be able to expect inputs or things available to run this + the ufs-weather-model on a platform other than Cheyenne, which will help in reviewing/testing the PR?  
* Should the CESMCOUPLED be W3_CESMCOUPLED? So it's just a "switch"? 

Again, thanks for all the awesome work and opening the draft PR.

model/src/w3updtmd.F90
DeniseWorthen: I'm not clear on what you're referring to as a switch here? In my switch_meshcap file I have an item "UWM" to add this line only for UWM or CESM. How should I have added it?

The line was added as part of testing for restart repro.

JessicaMeixner-NOAA: I'm wondering if instead of: 

```
#if defined(W3_UWM) || defined(CESMCOUPLED)
               charn(jsea) = aalpha
#endif      
```          

we should simply have: 
```
              charn(jsea) = aalpha
```

I'll run some tests to see the impact.

DeniseWorthen:
> 

I've completed the update of my tests for waves+gocart now. I need to do a final round of updates for the actual tests. Once I do that, I can create a staged input data directory on hera. I know certain files will still need modifications---the mod_defs especially. It should probably be w/in the next day or so that I can have something workable on hera.

model/src/w3updtmd.F90
DeniseWorthen: I was just trying to avoid modification of ww3 code for anything other than these two configurations.

model/src/w3wavemd.F90
```diff
@@ -1540,11 +1545,9 @@ SUBROUTINE W3WAVE ( IMOD, ODAT, TEND, STAMP, NO_OUT &
 !
 ! 3.3.3 Update ice floe diameter
 !
-#ifdef W3_IS2
+#if defined(W3_IS2) || defined(CESMCOUPLED)
```
JessicaMeixner-NOAA: Why is just the IS2 switch not sufficient here?

model/src/w3updtmd.F90
JessicaMeixner-NOAA: I'm testing this change: https://github.com/JessicaMeixner-NOAA/WW3/tree/bug/charnforrestart in the standalone WW3 regression tests as its my guess that this is a bug fix needed for all configurations.   I'll let you know how the tests go.

JessicaMeixner-NOAA: I ran some tests and we do get differences, which is not unaccepted necessarily.  However, we it's taking me longer than expected to make sure there's are the diffs we want.  For now we can keep this behind the switches here, but I think we can make this more general in the very near future.

DeniseWorthen: Thanks for testing; I'm behind on getting this and the UWM PR opened so I haven't addressed any comments yet. I needed a fix in the cap code and I've had trouble w/ the c192 test resources. I think I've resolved the issues and I'm testing hera and orion now.

JessicaMeixner-NOAA: Is the 1/4 deg grid going to change at all?  If not, if it could be copied to hera (or another machine other than Cheyenne) I could in parallel start creating ICs.  Apologies if that's already been done and I've missed that.

DeniseWorthen: There are two issues I think need to be addressed for the mod_def files. 

First, I am running w/o any obstr grid (they're all 0s).  At one point, Ali generated actual obstruction files to use when generating the mod_defs but I didn't incorporate them because I did not want to have to change the baseline I was testing against. 

Secondly, I created the grid.inp for the 1 deg tripole grid from the global 1-deg grid and for the 1/2 and 1/4 tripole grid from the 30m global grid. There are two things here which may need adjustment. First, the time-step settings and secondly the number of frequencies. I think for the actual prototypes, you may be using 36, but the RTs are using only 24 (?). 

The input-data on Hera is located at ``/scratch1/NCEPDEV/stmp4/Denise.Worthen/input-data-20220430/``. Within the ``WW3_input_data_20211113/createmoddefs`` are the 3 ww3_grid.inp.mx[100,050,025] that I have been using. 

Each of these lists a set of ascii files for input which I've generated from the fixed grid generation utility for the coupled model. These are all located at ``/scratch2/NCEPDEV/climate/Denise.Worthen/grids-esmf-20220116``

model/src/w3iogomd.F90
```diff
@@ -1562,8 +1616,39 @@ SUBROUTINE W3OUTG ( A, FLPART, FLOUTG, FLOUTG2 )
             USSCO=FKD*SIG(IK)*WN(IK,ISEA)*COSH(2.*KD)
             BHD(JSEA) = BHD(JSEA) +                             &
                 GRAV*WN(IK,ISEA) * EBD(IK,JSEA) / (SINH(2.*KD))
+#ifdef CESMCOUPLED
```
JessicaMeixner-NOAA: All of these extra outputs would be nice to be available to all users instead of just behind the CESMCOUPLED ifdefs.  They would need to be updated in the manual with definitions and the template files for input files would need to be added along with a regression test testing the new output to ensure that they are tested and are not unintentionally broken by other users.  I understand that time will likely not allow for this to be done immediately, so perhaps we can make an issue for this to be addressed at a future time.

model/src/w3iogoncdmd.F90
```diff
@@ -0,0 +1,759 @@
+#include "w3macros.h"
```
JessicaMeixner-NOAA, line 7: This is a really great feature.  I think this perhaps could inspire a larger conversation about how to optionally add in netcdf instead of binary output that also takes advantage of some of the recent work in ww3_ounf which allows for things like flexible user defined metadata and be usable by those who are not using CMEPS.

model/src/w3iorsmd.F90
```diff
@@ -430,18 +438,32 @@ SUBROUTINE W3IORS ( INXOUT, NDSR, DUMFPI, IMOD, FLRSTRT )
 !
 ! open file ---------------------------------------------------------- *
 !
+#ifdef CESMCOUPLED
```
JessicaMeixner-NOAA: Long term I don't think it's reasonable to support every systems custom naming conventions for restart files.  I hope that this can spur a conversation about how we could add some flexibility via namelist/input options that can add some customization meeting more users needs without system specific ifdefs.

model/src/w3sic4md.F90
```diff
@@ -388,8 +401,13 @@ SUBROUTINE W3SIC4 (A, DEPTH, CG, IX, IY, S, D)
       IF (INFLAGS2(-2)) ICECOEF6 = MUDD(IX,IY) ! a.k.a. MDN
       IF (INFLAGS2(-1)) ICECOEF7 = MUDT(IX,IY) ! a.k.a. MTH
       IF (INFLAGS2(0 )) ICECOEF8 = MUDV(IX,IY) ! a.k.a. MVS
-
+#ifdef CESMCOUPLED
+      ! It does not look like IC4PARS(1) is initialized anywhere - so
+      ! hard-wiring it here for now
+      IC4METHOD = 8
```
JessicaMeixner-NOAA: Based on the comment, perhaps this is a bug-fix that should be pulled out and tested without the #ifdef CESMCOUPLED unless IC4METHOD is a CESMFCOUPLED variable?

model/src/w3sic4md.F90
```diff
@@ -514,6 +532,50 @@ SUBROUTINE W3SIC4 (A, DEPTH, CG, IX, IY, S, D)
            END DO
            WN_I= 0.5 * ALPHA
 
+#ifdef CESMCOUPLED
+        CASE (8)
```
JessicaMeixner-NOAA: Is this something that could be added as an option that all can use?  This also seems like it could perhaps be a PR on it's own and is unrelated to the new cap although perhaps there is a dependency I do not fully understand here?

model/src/w3timemd.F90
```diff
@@ -77,6 +77,9 @@ MODULE W3TIMEMD
       INTEGER, PRIVATE        :: PRFTB(8)
       LOGICAL, PRIVATE        :: FLPROF = .FALSE.
       CHARACTER, PUBLIC       :: CALTYPE*8
+#ifdef CESMCOUPLED
+      LOGICAL, PUBLIC         :: NOLEAP = .TRUE.
```
JessicaMeixner-NOAA: Not sure why the updates here are CESM specific other than not wanting to modify the non-cap related code?

model/src/w3timemd.F90
```diff
@@ -541,13 +548,17 @@ INTEGER FUNCTION MYMD21 ( NYMD )
       NY   = NYMD / 10000
       NM   = MOD(NYMD,10000) / 100
       ND   = MOD(NYMD,100)
+#ifdef CESMCOUPLED
+      LEAP=.false. ! seems to be not working when I made NOLEAP=.true., still getting leap day
```
JessicaMeixner-NOAA: Is there a bug that needs to be reported and addressed related to this comment?

model/src/wav_shel_inp.F90
```diff
@@ -0,0 +1,1028 @@
...
+    call ESMF_UtilIOUnitGet(mds(13)); open(unit=mds(13) , status='scratch')
+    close(mds(5)); close(mds(6)); close(mds(7)); close(mds(8)); close(mds(9)); close(mds(10))
+    close(mds(11)); close(mds(12)); close(mds(13))
+
+    ntrace(1) = mds(3)
+    ntrace(2) = 10
+
+  end subroutine set_shel_io
+
+  subroutine set_shel_inp(dtime_sync)
```
JessicaMeixner-NOAA, line 106: Based on the below settings, can we provide a ww3_shel.inp (or in the future a ww3_shel.nml) file so that we do not have this hard-coded input section?   This would allow flexibility to for example add an additional variable to the output without having to recompile the code and just modify an input file.

model/src/wav_shel_inp.F90
```diff
@@ -0,0 +1,1028 @@
...
+    ! npts, pnames are fpr point output
+    allocate ( x(1), y(1), pnames(1) )
+    npts = 0
+    pnames(1) = ' '
+    prtfrm = .false.
+
+  end subroutine set_shel_inp
+
+  !===============================================================================
+  subroutine read_shel_inp(mpi_comm)
```
JessicaMeixner-NOAA, line 357: I don't think time will allow, but I'm thinking we should perhaps make this an independent subroutine that both ww3_shel.F90 would call as well the mesh based cap.  Additionally we can add in a future subroutine for ww3_shel.nml reading as well.  This way there would not be duplicative code.

JessicaMeixner-NOAA (commented):
@DeniseWorthen I know that you made the changes trying to minimize impacts to others which is appreciated.  However, I'm wondering if you have found a few bugs and added some great features that others would benefit from as well.  Due to time some of these items might have to become issues and addressed as a second step.

model/src/wav_shel_inp.F90
DeniseWorthen: I think breaking out the namelist/inp reading from the ww3_shel would make sense. Right now, what is in wav_shel_inp is probably too specific to the options used by UWM.

model/src/wav_shel_inp.F90
DeniseWorthen: The set_shel_inp is only used by CESM (it is called as part of waveinit_cesm).

JessicaMeixner-NOAA: Even if we are not using this, I still think it would be a nice feature for CESM to have that flexibility and would be cleaner code-wise.

model/src/w3iogomd.F90
DeniseWorthen: Agreed; this was meant to minimize any impact w/ the current code. Bringing in features like this for wider use would definitely be something we can work on.

model/src/w3timemd.F90
DeniseWorthen: I think the issue was that the CALTYPE is set during the creation of the mod_def. I'm not sure why it is there and not in an input namelist/inp file?

model/src/wav_comp_nuopc.F90
DeniseWorthen: It seems that the code documentation is being updated to utilize doxygen? If so, I think it makes more sense to move directly to inserting the required doxygen formatting.

JessicaMeixner-NOAA: We are not removing the WW3 headers as part of the switch to doxygen.  Doxygen at this time is an addition.  Adding doxygen would be a welcomed addition and perhaps it could be a compromise for not adding the WW3 headers.  Perhaps we could still update this header to have proper authorship and dates if others are not added? @MatthewMasarik-NOAA can help if you have any questions about doxygen formatting.  We are not requiring doxygen headers since we haven't added doxygen to the rest of the code yet, but it would certainly be very welcomed to add now instead of coming back later to add it in.

DeniseWorthen: I'm familiar w/ doxygen because I needed to add it for the grid generation code being added to ufs-utils. Does the documentation get built automatically for every push, as w/ ufs-utils?

MatthewMasarik-NOAA: @DeniseWorthen We have not hooked in the build automation yet.  I'm actually working right now on a getting a current draft PR with the first batch of marked-up files ready for submission.  This will happen later today or tomorrow.  After that we'll be in a position to set up the auto build.

model/src/w3iogomd.F90
DeniseWorthen: The calculations require the surface layer depth HSL, set as 0.2* HML, where HML is the mixed layer depth imported from the Ocean. We don't currently export this field from the Ocean in UWM so this was kept behind the cesm ifdef.

model/src/wav_comp_nuopc.F90
DeniseWorthen: I've started adding the doxygen documentation to a separate branch which I will merge in to this PR branch when I've got the majority of it done. I can build it by hijacking the documentation build in my ufs-utils fork. 

To start, I used the ``ww3_doxy_tmpl.md`` file for the required features. What I found though is that if add both ``@param`` and then also ``!< `` to document the variables, it ends up "double documenting". For example, this code snip:

```
!> Determine if a field is in a state
!!
!! @param[in] State    an ESMF_State
!! @param[in] fldname  the name of the ESMF field
!!
!> @author mvertens@ucar.edu, Denise.Worthen@noaa.gov  @<01-05-2022>
  logical function state_fldchk(State, fldname)
    ! ----------------------------------------------
    ! Determine if field is in state
    ! ----------------------------------------------

    ! input/output variables
    type(ESMF_State) , intent(in)  :: State                  !< ESMF_State object
    character(len=*) , intent(in)  :: fldname                !< a field name
```

produces the following
Screen Shot 2022-04-05 at 8 08 10 AM

I wonder if it makes more sense to use @param only (or vice-versa) since documenting the arguments twice is redundant.

MatthewMasarik-NOAA: Hi @DeniseWorthen , yes you're right, the two should not be used together. The `@param` should be used in headers for subroutines/functions, and the `!<` just for inline documenting of module variables.  All module variables are supposed to be documented.

MatthewMasarik-NOAA: kudos for adding the doxygen markup!

DeniseWorthen: @MatthewMasarik-NOAA Regarding module variables, do variables which are local only need to be docmented? That would mean things like ``integer :: i,j`` ?

MatthewMasarik-NOAA: @DeniseWorthen , variables local to subroutines and functions don't need to be documented. Just variables that are listed between the 'module' and 'contains' keywords.  Does that answer it?

MatthewMasarik-NOAA: When I said 'all' above, that's a little misleading ;)  only those non-locals.

model/src/w3iogomd.F90
JessicaMeixner-NOAA: I'll have to look more but it seems like the "surface stokes drift" doesn't depend on HSL and we actually already have a surface stokes drift calculation (USSX,USSY) so understanding how those are different will be important at some point.  We might should add HML as an import field in the standalone mode to WW3 then?

model/src/w3iorsmd.F90
JessicaMeixner-NOAA: Would it be possible to find out the requirements for a resatrt file name for CESM so we can start a discussion about how to accommodate CESM in a more general way?

model/src/w3wavemd.F90
mvertens: @Jessica - CESM does not use W3_IS2 - and I did not write this code. I think this is a question for Cecelia Bitz or Lettie Roach who put this into the current ww3 code. @alperaltuntas - do you know the answer to this?

model/src/w3iogomd.F90
mvertens: @JessicaMeixner-NOAA - again I think this is a question for CC and Lettie. I'm not sure what their github ids are. @alperaltuntas - are you working on this particular part of the code base?

model/src/w3sic4md.F90
mvertens: @JessicaMeixner-NOAA - I'm happy to try to set up a call with CC,Lettie and @alperaltuntas  if that would help.
Please let me know.

model/src/w3timemd.F90
mvertens: @JessicaMeixner-NOAA - I think we can get rid of this in the CESM initialization.

model/src/w3sic4md.F90
JessicaMeixner-NOAA: @mvertens It would be great to get some answers to some of these questions.  If that would be easier in a call, then that would be great. Otherwise, I'm happy to continue via the GitHub interface.

model/src/w3wavemd.F90
alperaltuntas: I am not sure. I am mentioning @lettie-roach who may know the answer.

model/src/w3iogoncdmd.F90
DeniseWorthen: I currently have a feature branch which updates w3iogoncmd to write a netcdf file for any variable requested in the inp file. This current version was only set up to write a limited number of fields.

model/src/w3iogomd.F90
alperaltuntas: These are the stokes drift calculations added by @qingli411, which are used in CESM's legacy OCN-WAV coupling that is based on passing the computed stokes drifts and langmuir enhancement factor to CVMix (through POP2 or MOM6). These changes were initially introduced to NCAR's old WW3 fork based on v.3.14.  @qingli411 can elaborate more.

model/src/w3sic4md.F90
mvertens: @lettie-roach - can you answer the above question from Jessica?

model/src/w3sic4md.F90
mvertens: @lettie-roach - can you address this?

model/src/w3iogomd.F90
qingli411: These Stokes drift calculations were added to enable the Langmuir turbulence parameterization in CESM through CVMix ([Li et al., 2016](https://doi.org/10.1016/j.ocemod.2015.07.020); [Li and Fox-Kemper 2007](https://doi.org/10.1175/JPO-D-17-0085.1)), which requires some scalar quantities such as the surface layer averaged Langmuir number and the Langmuir multiplier. These quantities can be diagnosed from the full profile of the Stokes drift. But they were added here in WW3 because (1) the full profile of Stokes drift is an array with depth rather than a scalar which is more expensive to pass from WW3 to the ocean model, and (2) the vertical resolution of the ocean model is too coarse to resolve the shape of the Stokes drift profile. 

@JessicaMeixner-NOAA The surface Stokes drift doesn't depend on the surface layer depth HSL (defined as 1/5 the mixed layer depth HML). It is calculated in newer versions of WW3 but was not there in the old version we had in CESM based on v3.14. Actually I think I ported this specific feature from v4.18. A difference in the calculation is that we added the contribution from a f^{-5} spectral tail, which seems to be commented out in newer versions of WW3. The surface layer averaged Stokes drift depends on HSL, but I'm not sure if it makes sense to add it as a standard WW3 output -- as you mentioned we will need to add HML as an import field for that but HML is not used in the calculation of any prognostic quantities in WW3. I think right now the use of these variables and the way they were implemented are really specific to CESM (or CVMix to be more accurate). Ideally, it would be nice to have a standard and easy way to pass the full Stokes drift profile to the ocean. Then these quantities can be diagnosed in the ocean model.

model/src/w3sic4md.F90
lettie-roach: Hi @JessicaMeixner-NOAA -  I put this in because I couldn't figure out how to set this variable properly. I think it should be a namelist variable - it's not specific to CESMCOUPLED

model/src/w3sic4md.F90
lettie-roach: We added case 8 as a new option for the attenuation scheme, so yes it's unrelated to the new cap. We have more changes to make to the attenuation scheme based on some recent results, so it would make sense to be a PR on its own

model/src/w3timemd.F90
lettie-roach: I think Cecilia put in the leap-year related changes to fix a bug, so not CESM specific

model/src/w3wavemd.F90
lettie-roach: I don't think we need CESMCOUPLED here

model/src/w3timemd.F90
JessicaMeixner-NOAA: @cmbitz can you confirm if this should be pulled out a bug-fix and not something CESM specific?

model/src/w3iogomd.F90
JessicaMeixner-NOAA: @qingli411 if this does not make sense as an output of WW3 perhaps it should be moved to just a calculation of an import/export in the nuopc cap instead of being in the output computation routine?

model/src/w3timemd.F90
DeniseWorthen: I think this and the change at line 551 (below) are because the calendar type is initialized in the mod_def file instead of a namelist. I believe we can resolve this issue in the same way @mvertens resolved changing the timestep after the model initialization in the ``waveinit_cesm`` routine. But we will need to test and verify it works before removing this particular ifdef.

model/src/w3iogomd.F90
qingli411: @JessicaMeixner-NOAA I think it does make sense to have these variables as output for CESM -- when running CESM coupled with WW3, it would be good to have some of them in the output. But I'm just not sure if it makes sense to add them to the standard output of WW3 if other users do not use them... So, perhaps it is OK to keep them behind the `CESMCOUPLED` ifdefs for now? In the future, I think it would be good to have a more general way to pass the full Stokes drift profile information to the ocean, e.g., something like what @breichl did? Then these quantities can be diagnosed from the Stokes drift profile in the ocean model.

JessicaMeixner-NOAA: I think that's what's at issue here, we'd like for features that are not isolated to the cap to not be CESM only features and be available for non-CESM users.  If that requires adding an optional input as well, then we should add the optional input and make checks that say if you don't have x input then you can't have y output, as required.

JessicaMeixner-NOAA: That being said, if this is temporary and we make an issue to remove this as you move to use the full Stokes profile, then perhaps just making an issue is fine.  However, then we have the issue that these are taking up places in the output array structures and so the next person who adds a variable is going to add something after that and then that gets messy to reconcile, so maybe this should use different indices so it'd be easier for others to add new variables in the meantime?

model/src/w3sic4md.F90
JessicaMeixner-NOAA: Have you tried setting in ww3_grid.inp: 
  &SIC4 IC4METHOD = 8 / 
?

model/src/w3iogomd.F90
qingli411: @JessicaMeixner-NOAA OK I see. How about this? We can add a new standard output in WW3 for a depth averaged Stokes drift which requires a new import field of a depth. We don't need to specify the meaning of this depth, i.e., it can be 1/5 of the mixed layer depth as we've been using in CESM, but it can also be any depth of interest for other users. In the limit that this depth approaches zero, the depth averaged Stokes drift should be equivalent to the surface Stokes drift. In the limit that this depth approaches the water depth, the depth averaged Stokes drift should be the total Stokes transport divided by the water depth. I think both surface Stokes drift and total Stokes transport are standard output of WW3. This depth averaged Stokes drift can serve as another variable to describe the Stokes drift, which is potentially useful to other users. Then, I think the calculation of all other variables such as the Langmuir multiplier can be moved to the nuopc cap. What do you think?

There is another issue that whether we should include the contribution of a spectral tail to the Stokes drift. We are including the contribution from a f^{-5} spectral tail when using it in CESM. But I suspect that the Langmuir multiplier might not be very sensitive to this choice, so the effect of dropping this contribution is probably small...

JessicaMeixner-NOAA: @qingli411 that sounds good to me.  

In terms of the tail to the Stokes drift, I'd recommend making an issue and then creating a flag to add it or not in a separate PR. My guess is others will be just as interested as you to have the option to add the tail or not.  But this way we only have one calculation of surface Stokes drift instead of multiple.

model/src/w3sic4md.F90
lettie-roach: Aha, no! We didn't have that file in the version we were using (https://github.com/lettie-roach/WW3-CESM/tree/lrmod, forked from Helen Kershaw). That is how it should be set

model/src/w3iogomd.F90
qingli411: @JessicaMeixner-NOAA Sounds good!

JessicaMeixner-NOAA:
@DeniseWorthen in the wmesmf cap, the inputs are either from "cpl" or you can use the original method of inputs.  It looks like if we're using ww3_shel.inp that all inputs must come from the coupler?   Or is there another way that is determined?

DeniseWorthen:
@JessicaMeixner-NOAA I'm not clear on your question. Are you asking if the mesh cap can be run standalone or when the forcing fields come from files?

JessicaMeixner-NOAA:
> @JessicaMeixner-NOAA I'm not clear on your question. Are you asking if the mesh cap can be run standalone or when the forcing fields come from files?

So let's say we wanted to coupled atm-wav, so that winds are coming from the coupler, but we'd like to have an input field of ice that comes from a file and we are using ww3_shel.inp.  How would that get specified? Would that even be possible?

mvertens:
@JessicaMeixner-NOAA - with CDEPS inline functionality (so that a prognostic component can ingest forcing fields like you want and easily from its cap) - this could be easily done today. I'm happy to show how this can be implemented. I can go over this briefly at our upcoming call next week.

JessicaMeixner-NOAA:
> @JessicaMeixner-NOAA - with CDEPS inline functionality (so that a prognostic component can ingest forcing fields like you want and easily from its cap) - this could be easily done today. I'm happy to show how this can be implemented. I can go over this briefly at our upcoming call next week.

This is a capability we have with wmesmf when using ww3_multi.inp without using a separate data model.

mvertens:
@JessicaMeixner-NOAA - you are not using a separate data model - just share code functionality. I'm not suggesting this is the only way - but a powerful way to have different spatial and time interpolations (including 3d mapping) for forcing fields that it would be helpful for everyone to be aware of.

model/src/w3initmd.F90
JessicaMeixner-NOAA: Not creating issue for this, assuming it will be taken care of in this PR.

model/src/w3initmd.F90
```diff
@@ -1402,6 +1415,16 @@ SUBROUTINE W3INIT ( IMOD, IsMulti, FEXT, MDS, MTRACE, ODAT      &
           CALL ALL_VA_INTEGRAL_PRINT(IMOD, "W3INIT, step 8.1")
 #endif
 #endif
+#ifdef CESMCOUPLED
```
JessicaMeixner-NOAA: Looking through this again, this seems like debug comments that are not needed long term.  Could this be cleaned out as well?

model/src/w3iogomd.F90
JessicaMeixner-NOAA: The proposed subsequent updates are covered in these issues: 
* Add tail to USSX/USSY and do not add extra variable: https://github.com/NOAA-EMC/WW3/issues/669 
* Add "depth" input field. https://github.com/NOAA-EMC/WW3/issues/670 
* Move any relevant variable calculations to the cap code that do not need output: https://github.com/NOAA-EMC/WW3/issues/671
* Remove the ifdefs for any other remaining variables so that all users can access new output variables https://github.com/NOAA-EMC/WW3/issues/672

model/src/w3iogoncdmd.F90
JessicaMeixner-NOAA: An issue was created for this: https://github.com/NOAA-EMC/WW3/issues/673 where we can continue this discussion

model/src/w3iorsmd.F90
JessicaMeixner-NOAA: Issue to address this is here: https://github.com/NOAA-EMC/WW3/issues/674

model/src/w3sic4md.F90
JessicaMeixner-NOAA: An issue for this is created here: https://github.com/NOAA-EMC/WW3/issues/675

model/src/w3sic4md.F90
JessicaMeixner-NOAA: An issue to remove ifdefs is here: https://github.com/NOAA-EMC/WW3/issues/676
I will leave the creation of a new issue about the additional changes to others.

model/src/w3timemd.F90
JessicaMeixner-NOAA: I have not created an issue for this as I believe it will be taken care of in this PR.

model/src/w3updtmd.F90
JessicaMeixner-NOAA: The follow up for this issue is here: https://github.com/NOAA-EMC/WW3/issues/677

model/src/w3wavemd.F90
JessicaMeixner-NOAA: Assuming this will be updated in this PR and so no issues is made.

model/src/wav_shel_inp.F90
JessicaMeixner-NOAA: Follow up issue: https://github.com/NOAA-EMC/WW3/issues/678

model/src/wav_shel_inp.F90
JessicaMeixner-NOAA: Issue for this is here: https://github.com/NOAA-EMC/WW3/issues/679

JessicaMeixner-NOAA (changes requested):
Issues to address review comments have been made and linked in the appropriate places.  In addition items that I believe will be taken care of in this PR were also noted.  If these items will not be addressed in this PR please let me know so an issue for that item can be created.

model/src/w3wavemd.F90
DeniseWorthen: @JessicaMeixner-NOAA The combination of W3_IS2 and CESMCOUPLED has been removed.

model/src/w3sic4md.F90
DeniseWorthen: @JessicaMeixner-NOAA The line in question for hardcoding the ic4method has been removed.

model/src/w3timemd.F90
DeniseWorthen: @JessicaMeixner-NOAA The CESMCOUPLED ifdefs to hardcode noleap have been removed.

model/src/w3initmd.F90
DeniseWorthen: @JessicaMeixner-NOAA I have removed some sections where debugging print statements were added for CESMCOUPLED.

JessicaMeixner-NOAA (commented):
All issues that were going to be taken care of before the PR was merged have been addressed.

JessicaMeixner-NOAA (approved):
The regression tests are as expected: 

```
mww3_test_03/./work_PR2_UNO_MPI_d2                     (15 files differ)
mww3_test_03/./work_PR1_MPI_d2                     (16 files differ)
mww3_test_03/./work_PR3_UNO_MPI_d2_c                     (14 files differ)
mww3_test_03/./work_PR3_UQ_MPI_d2_c                     (15 files differ)
mww3_test_03/./work_PR3_UNO_MPI_d2                     (12 files differ)
mww3_test_03/./work_PR3_UQ_MPI_d2                     (16 files differ)
ww3_ta1/./work_UPD0F_U                     (0 files differ)
ww3_tp2.17/./work_ma                     (1 files differ)
ww3_tp2.17/./work_a                     (1 files differ)
ww3_tp2.17/./work_mb                     (1 files differ)
ww3_tp2.17/./work_mc                     (1 files differ)
ww3_tp2.17/./work_c                     (1 files differ)
ww3_tp2.17/./work_b                     (1 files differ)
ww3_tp2.6/./work_ST0                     (1 files differ)
ww3_tp2.6/./work_ST4                     (1 files differ)
ww3_tp2.6/./work_pdlib                     (1 files differ)
ww3_ufs1.3/./work_a                     (1 files differ)
```
There are the known b4b cases + the unstructured mod_def files which have at times been not b4b lately.

PR #680, opened 2022-04-21
Add convert='big_endian' to each unformatted open.

### LINKED ISSUE #668: Add convert=big_endian to any open statement of binary files
**Is your feature request related to a problem? Please describe.**
With GNU compilers, the main program set's the "endian" to big or little.  Therefore if WW3 is compiled as a library used in a coupled system that is setting (or not setting) the endian-ness then the main program will dictate this.  This issue was run into in the ufs-weather-model, see issue https://github.com/ufs-community/ufs-weather-model/issues/1115

**Describe the solution you'd like**
Specify on all binary open statements "convert=big_endian"

**Describe alternatives you've considered**
We could default to whatever the top-level coupled exe or system is using, but this leaves ambiguity and could bring other problems and complications as most WW3 applications are currently using big_endian and this would likely break things unexpectedly. 

**Additional context**

kgerheiser:
Explicitly use `convert='big_endian'` option instead of relying on compiler flags. -fconvert=big_endian on GNU is only applied when compiling the main program. No result differences expected.

This causes conflicts when building UFS coupled with WW3.

See https://github.com/ufs-community/ufs-weather-model/issues/1115

Fix #668

This is non-standard Fortran, but it is supported by all compilers that I checked and removes the need to set the option for each compiler.

[Intel](https://www.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/language-reference/file-operation-i-o-statements/open-statement-specifiers/open-convert-specifier.html#open-convert-specifier)

[GNU](https://gcc.gnu.org/onlinedocs/gfortran/CONVERT-specifier.html)

[PGI](https://www.pgroup.com/resources/docs/19.10/x86/fortran-ref-guide/index.htm#fort-io-open-close-files)

[Cray](https://support.hpe.com/hpesc/public/docDisplay?docId=a00115296en_us&page=Input_Output_Statements.html)

[Flang/F18](https://flang.llvm.org/docs/Extensions.html)

[NAG](https://www.nag.com/nagware/np/r62_doc/manual/compiler_5_1.html)

[IBM](https://www.ibm.com/docs/en/xl-fortran-aix/15.1.2?topic=attributes-open)

[Lahey](http://www.lahey.com/doc.htm)

### Check list  

- [x] Branch is up to date with the authoritative repository (NOAA-EMC) develop branch. 
- [x] Checked the [checklist for a developer submitting to develop](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-a-developer-submitting-to-develop). 
- [x] If a version number update is required, checked the [updating version number](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-updating-version-number) checklist. 
- [x] If a new feature was added, a regression test for testing the new feature is added. 
 
### Testing

Run full set of regression tests on Hera. And GCC + Intel Github Action builds pass.

commit message:
Add endianness option (-DENDIAN=BIG/LITTLE/NATIVE)


JessicaMeixner-NOAA:
@kgerheiser thank you for providing this code fix!

aliabdolali (commented):
@kgerheiser I ran the regtests and noticed we have `current.ww3`, `wind.ww3`,` ice.ww3` ... different. These binary files are generated after `ww3_prep` and `ww3_prnc` execution. I looked at `ww3_bounc` and `ww3_bound` which generate `nest.ww3` but they look ok. 

[matrixCompFull.txt](https://github.com/NOAA-EMC/WW3/files/8563412/matrixCompFull.txt)
[matrixCompSummary.txt](https://github.com/NOAA-EMC/WW3/files/8563414/matrixCompSummary.txt)
see the diff in binary files:
binary_comp

kgerheiser:
Thank you for bringing this up @aliabdolali. The data looks the same but ordered differently (big endian vs. little endian), probably because of the compiler flag being removed. Maybe I missed a convert somewhere.

aliabdolali:
> Thank you for bringing this up @aliabdolali. The data looks the same but ordered differently (big endian vs. little endian), probably because of the compiler flag being removed. Maybe I missed a convert somewhere.

@kgerheiser I agree, it is flipped left<->right. My initial thought was we could just compare the changes in ww3_bounc/bound which is ok vs ww3_prnc/prep, what do you think?

aliabdolali:
Hi @kgerheiser any update on this? Jun and ufs-weather-model team are waiting for an estimated merge date for this PR.

kgerheiser:
Found a couple issues. I should be able to have a fix in today.

aliabdolali:
> Found a couple issues. I should be able to have a fix in today.

amazing, let me know if you need a check from my end.

kgerheiser:
@aliabdolali I just pushed a commit. I went through each open pretty finely, and think I have it all fixed. Could you help me test this? I can run the regression tests, but didn't figure out the compare script or running individual tests.

aliabdolali:
> @aliabdolali I just pushed a commit. I went through each open pretty finely, and think I have it all fixed. Could you help me test this? I can run the regression tests, but didn't figure out the compare script or running individual tests.

I will check it today.

aliabdolali:
@kgerheiser 
I tested the changes, it looks fine for all except a few unstructured cases. I did not have a chance to dig more as Hera went under maintenance. Could you check if you can run ww3_tp2.17 test? or ww3_tp2.21? The model compiles but failed during execution. 
I will dig more once Hera is back,

ukmo-ccbunney:
Whilst I am not against this change, it will break my Operational Suite (which is currently using native [little] endian I/O).

I am happy to have `CONVERT` statements in all the unformatted I/O - but perhaps we could pass in a string for the endianness that is set to "big_endian" by default but can be set by the user at runtime (i.e. to "native", "big_endian" or "little_endian"), rather than hard-coding it?

Edit: Thinking about this, I can see that the problem with that approach is that every program would need to have a way of specifying the endianness at runtime. Could something be added to the CMake config to set an `ENDIAN` value in constants.F90 at  the configuration step?

kgerheiser:
@ukmo-ccbunney, I was having the same thought about making  variable as I made these changes just to avoid duplication because there are so many instances.

Now that I understand the issue a little better now. There doesn't seem to be a great way to handle this. 

The `-fconvert=big_endian` flag for GNU only applies to the main executable and this has a conflict when running coupled with the UFS where another component conflicts. So, it needs to be explicit and every instance of the open needs it, which is a lot of places (29 files changed). And as evidenced here it can be tricky to track down. It also adds the burden of making sure new file opens also adhere to this rule. A compiler flag is simpler in that sense.

For the time being the CMake idea is good. What about handling it with runtime configuration?

aliabdolali:
@kgerheiser
If we step back and look at it again:
if we do not specify endianness, does the machine and other components (for coupled cases) mandate little or big endian? 
I understand if we remove it across the code, we will have changes in all regtests and also need to regenerate many files, but that's a one-time thing. Then it will become more general?
P.S. it is the outcomes of a conversation I had with Arun today.

kgerheiser:
As noted by @junwang-noaa, we came to the conclusion that it would be easiest to continue with this PR. So, I created a CMake option (`-DENDIAN=BIG/LITTLE/NATIVE`, default is `BIG`) and changed the hardcoded endianness to a string variable in `constants.F90` which is set by a compile time definition.

Does this approach work?

aliabdolali:
> As noted by @junwang-noaa, we came to the conclusion that it would be easiest to continue with this PR. So, I created a CMake option (`-DENDIAN=BIG/LITTLE/NATIVE`, default is `BIG`) and changed the hardcoded endianness to a string variable in `constants.F90` which is set by a compile time definition.
> 
> Does this approach work?

I like the approach, it is giving us flexiblity. 
I tried to test a regtests, so I added `-DENDIAN="BIG"` to `CMAKE_OPTIONS` in regtests/bin/run_cmake_test, but it failed.

`cmake $path_cmake ${CMAKE_OPTIONS} -DENDIAN="BIG" -DSWITCH=$path_build/switch -DCMAKE_INSTALL_PREFIX=install > $ofile 2>&1`

error is
`model/src/w3strkmd.F90(545): error #7589: Not a valid value for the char-expr in this connect-spec.   ['W3_ENDIANNESS']`

Am I missing sth?

kgerheiser:
@aliabdolali do you have the latest commit (e509d8d)? I fixed the error you're seeing and there are no instances of `W3_ENDIANNESS` anymore. Then Intel pre-processor does not allow `-Dname=value` with an underscore in the name. It's one word now and not a valid string variable before pre-processing to make any break more obvious.

And FYI this will break the Makefile build. Does there need to be a corresponding change there?

aliabdolali:
> @aliabdolali do you have the latest commit ([e509d8d](https://github.com/NOAA-EMC/WW3/commit/e509d8d6895d98c68da4e35d065855432933fdd5))? I fixed the error you're seeing and there are no instances of `W3_ENDIANNESS` anymore. Then Intel pre-processor does not allow `-Dname=value` with an underscore in the name. It's one word now and not a valid string variable before pre-processing to make any break more obvious.
> 
> And FYI this will break the Makefile build. Does there need to be a corresponding change there?

I pulled it and tested it. I ran it once with DENDIAN="BIG" and once with DENDIAN="LITTLE", and then compared. Binary files are identical, is it expected? I assume not. Did I do something wrong? Could you give me a example cmake command?

kgerheiser:
No, I would not expect that.

Did you use the same build directory multiple times? If you re-ran CMake from the same build directory and just changed the variable `-DENDIAN=` then it'll still have the old value because CMake keeps it as a cache variable which is not overwritten. It's a weird quirk of CMake, but if you start with a fresh build it'll take the correct value.

Testing the build with different values I can see the variable filled correctly.

aliabdolali:
> No, I would not expect that.
> 
> Did you use the same build directory multiple times? If you re-ran CMake from the same build directory and just changed the variable `-DENDIAN=` then it'll still have the old value because CMake keeps it as a cache variable which is not overwritten. It's a weird quirk of CMake, but if you start with a fresh build it'll take the correct value.
> 
> Testing the build with different values I can see the variable filled correctly.

I have a different build dir for each run and therefore it should not be the same option.

Here is the step I took on Hera:

 ```
 module purge
  module load cmake/3.20.1
  module use /scratch2/NCEPDEV/nwprod/hpc-stack/libs/hpc-stack/modulefiles/stack
  module load hpc/1.1.0
  module load hpc-intel/18.0.5.274
  module load hpc-impi/2018.0.4
  module load netcdf/4.7.4
  module load jasper/2.0.25
  module load zlib/1.2.11
  module load png/1.6.35
  module load hdf5/1.10.6
  module load bacio/2.4.1
  module load g2/3.4.2
  module load w3nco/2.4.1
  module load esmf/8_2_0


```

**Run1)**
in regtests/bin/run_cmake_test

`cmake $path_cmake ${CMAKE_OPTIONS} -DENDIAN="BIG" -DSWITCH=$path_build/switch -DCMAKE_INSTALL_PREFIX=install > $ofile 2>&1`

./bin/run_cmake_test -o both -c ukmo_cray_gnu -S -s PR2_UNO -w **work_PR2_UNO_a** -m grdset_a ../model mww3_test_02



**Run2)**
in regtests/bin/run_cmake_test

`cmake $path_cmake ${CMAKE_OPTIONS} -DENDIAN="LITTLE" -DSWITCH=$path_build/switch -DCMAKE_INSTALL_PREFIX=install > $ofile 2>&1`

./bin/run_cmake_test -o both -c ukmo_cray_gnu -S -s PR2_UNO -w **work_PR2_UNO_b** -m grdset_a ../model mww3_test_02


So there are two build dir under `mww3_test_02/work_PR2_UNO_a` and `mww3_test_02/work_PR2_UNO_b`

**Comment by kgerheiser** [comment]
Do you have build logs or a path to look at for both the source and build?

In the compiler output you should see if the definition `-DENDIANNESS` is being passed to the compiler. Did you change all instances of `cmake $path_cmake`? I see 3 instances in `run_cmake_test`, or try passing it through `CMAKE_OPTIONS` so it's picked up by all.

You can also look at the literal pre-processed source by going into the build directory, then `model/src`, and run `make constants.i` and it'll create a `constants.i` which will have the actual value.

I did a standalone build on Hera and see the correct value being filled in, and I just submitted a job to replicate your run.

**Comment by kgerheiser** [comment]
I used a matrix like this and ran the same test twice:

```
export CMAKE_OPTIONS="-DENDIAN=BIG"
./bin/run_cmake_test -b slurm -o all -c hera.intel -S -T -s PR2_UNO -w work_PR2_UNO_big_endian -m grdset_a ../model mww3_test_02
export CMAKE_OPTIONS="-DENDIAN=LITTLE"
./bin/run_cmake_test -b slurm -o all -c hera.intel -S -T -s PR2_UNO -w work_PR2_UNO_little_endian -m grdset_a ../model mww3_test_02
```

The build logs look good, and I'm not sure which files to look at but I ran `shasum` on some output files and they were different.

You can see the build output here: 

```
/scratch1/NCEPDEV/nems/Kyle.Gerheiser/WW3/regtests/mww3_test_02/work_PR2_UNO_big_endian
/scratch1/NCEPDEV/nems/Kyle.Gerheiser/WW3/regtests/mww3_test_02/work_PR2_UNO_little_endian
```

**Comment by aliabdolali** [comment]
Awesome, I will do the same and will get back to you. You should compare out_grd.ww3, out_pnt.ww3, wind.ww3,... which are binary and different when we change from big to little. I'll let you know once i did a test. Thanks for testing it at your end

**Comment by junwang-noaa** [comment]
@aliabdolali May I ask if you are able to build and run big_endian and little_endian tests and get correct results? We are still waiting for this PR for the mesh cap commit. Thank you!

**Comment by aliabdolali** [comment]
> @aliabdolali May I ask if you are able to build and run big_endian and little_endian tests and get correct results? We are still waiting for this PR for the mesh cap commit. Thank you!


I am putting together the list of failed tests (standalone WW3) to share with @kgerheiser early tomorrow morning, so we can solve those.

File: model/src/w3fldsmd.F90 [hunk 1]
```diff
@@ -359,13 +359,13 @@ SUBROUTINE W3FLDO ( INXOUT, IDFLD, NDS, NDST, NDSE, NX, NY,     &
...
 #endif
-              OPEN (NDS,FILE=FPRE//FNAME(:I),FORM=FORM,ERR=803,       &
+              OPEN (NDS,FILE=FPRE//FNAME(:I),FORM=FORM,convert='big_endian',ERR=803,       &
                     IOSTAT=IERR)
           ELSE
 #ifdef W3_DEBUGFLS
        WRITE(740+IAPROC,*) '2 : W3FLDSMD FNAME=', FNAME(:I)
 #endif
-              OPEN (NDS,FILE=FNAME(:I),FORM=FORM,ERR=803,IOSTAT=IERR)
+              OPEN (NDS,FILE=FNAME(:I),FORM=FORM,convert='big_endian',ERR=803,IOSTAT=IERR)
```
aliabdolali: @kgerheiser should it be file_endian?

model/src/w3fldsmd.F90
```diff
@@ -359,13 +359,13 @@ SUBROUTINE W3FLDO ( INXOUT, IDFLD, NDS, NDST, NDSE, NX, NY,     &
 #ifdef W3_DEBUGFLS
        WRITE(740+IAPROC,*) '1 : W3FLDSMD FNAME=', FNAME(:I)
 #endif
-              OPEN (NDS,FILE=FPRE//FNAME(:I),FORM=FORM,ERR=803,       &
+              OPEN (NDS,FILE=FPRE//FNAME(:I),FORM=FORM,convert='big_endian',ERR=803,       &
```
aliabdolali: @kgerheiser should it be file_endian?

model/src/w3iogrmd.F90
```diff
@@ -551,10 +551,10 @@ SUBROUTINE W3IOGR ( INXOUT, NDSM, IMOD, FEXT )
 !AR: ADD DEBUGFLAG      WRITE(*,*) 'FILE=', FNMPRE(:J)//'mod_def.'//FILEXT(:I)
       IF ( WRITE ) THEN
           OPEN (NDSM,FILE=FNMPRE(:J)//'mod_def.'//FILEXT(:I),         &
-                FORM='UNFORMATTED',ERR=800,IOSTAT=IERR)
+                form='UNFORMATTED', convert='big_endian',ERR=800,IOSTAT=IERR)
```
aliabdolali: @kgerheiser should it be file_endian?

model/src/w3iogrmd.F90
```diff
@@ -551,10 +551,10 @@ SUBROUTINE W3IOGR ( INXOUT, NDSM, IMOD, FEXT )
...
         ELSE
           OPEN (NDSM,FILE=FNMPRE(:J)//'mod_def.'//FILEXT(:I),         &
-                FORM='UNFORMATTED',STATUS='OLD',ERR=800,IOSTAT=IERR)
+                form='UNFORMATTED', convert='big_endian',STATUS='OLD',ERR=800,IOSTAT=IERR)
```
aliabdolali: @kgerheiser should it be file_endian?

aliabdolali (changes requested):
@kgerheiser I still have some cases that failed, and I suspect some of the big_endians are still in the code. 
I ran two sets of simulations, one with big and one with little, and for both, I have some failed tests.

junwang-noaa:
@aliabdolali Can you check if your tests pass if you make the code changes as you suggested? Just want to check if that will fix the issue. Thanks

kgerheiser:
Took care of the hardcoded `big_endian` and found one missing unformatted open.

aliabdolali:
@kgerheiser The model compiles but there is an inconsistency in the grids
please try this with BIG and LITTLE.
`./bin/run_cmake_test -b slurm -o all -S -T -s PR1 -w work_PR1 ../model ww3_tp1.6`

And here is the error during `ww3_shel` execution. 
```
 Initializations :
 --------------------------------------------------
       Preparing input files ...
            water levels : file not needed

 *** WAVEWATCH III ERROR IN W3FLDO : 
     INCOMPATIBLE GRID DATA :   369098752   50331648   16777216
                  SHOULD BE :          22          3          1

```

aliabdolali:
@kgerheiser The problem is in model/src/w3fldsmd.F90, I reverted your commit [4cf86ff](https://github.com/NOAA-EMC/WW3/pull/680/commits/4cf86ffe53ab09d05d7e54c2777e145b28cdf1b9) back and it worked at least for the following test. i am not sure if it is the right way forward and if it fails other tests.

model/src/w3fldsmd.F90
```diff
@@ -203,6 +203,7 @@ SUBROUTINE W3FLDO ( INXOUT, IDFLD, NDS, NDST, NDSE, NX, NY,     &
 #ifdef W3_DEBUGFLS
       USE W3ODATMD, only : IAPROC
 #endif
+      use constants, only: file_endian
```
aliabdolali: I just removed this line + two other additions in model/src/w3fldsmd.F90 and tested regtests with both little and big endians. It seems the fix from the finished jobs. I'll let you know once all tests are done, so you can revert them.

model/src/w3fldsmd.F90
```diff
@@ -359,13 +360,13 @@ SUBROUTINE W3FLDO ( INXOUT, IDFLD, NDS, NDST, NDSE, NX, NY,     &
 #ifdef W3_DEBUGFLS
        WRITE(740+IAPROC,*) '1 : W3FLDSMD FNAME=', FNAME(:I)
 #endif
-              OPEN (NDS,FILE=FPRE//FNAME(:I),FORM=FORM,ERR=803,       &
+              OPEN (NDS,FILE=FPRE//FNAME(:I),FORM=FORM,convert=file_endian,ERR=803,       &
```
aliabdolali: I just removed this line + two other additions in model/src/w3fldsmd.F90 and tested regtests with both little and big endians. It seems the fix from the finished jobs. I'll let you know once all tests are done, so you can revert them.

model/src/w3fldsmd.F90
```diff
@@ -359,13 +360,13 @@ SUBROUTINE W3FLDO ( INXOUT, IDFLD, NDS, NDST, NDSE, NX, NY,     &
...
                     IOSTAT=IERR)
           ELSE
 #ifdef W3_DEBUGFLS
        WRITE(740+IAPROC,*) '2 : W3FLDSMD FNAME=', FNAME(:I)
 #endif
-              OPEN (NDS,FILE=FNAME(:I),FORM=FORM,ERR=803,IOSTAT=IERR)
+              OPEN (NDS,FILE=FNAME(:I),FORM=FORM,convert=file_endian,ERR=803,IOSTAT=IERR)
```
aliabdolali: I just removed this line + two other additions in model/src/w3fldsmd.F90 and tested regtests with both little and big endians. It seems the fix from the finished jobs. I'll let you know once all tests are done, so you can revert them.

aliabdolali:
@kgerheiser I can confirm reverting [4cf86ff](https://github.com/NOAA-EMC/WW3/pull/680/commits/4cf86ffe53ab09d05d7e54c2777e145b28cdf1b9)solved the problem for both BIG and LITTLE. I am doing the comparison with the base. Could you make the changes and push?

aliabdolali:
@thesser1 @maryanderson64 @ukmo-ccbunney @mickaelaccensi @sbrus89 
Here, we added an option for endianness, so users can define it during the build and select one of BIG, LITTLE and NATIVE options.
export CMAKE_OPTIONS="-DENDIAN=BIG"
export CMAKE_OPTIONS="-DENDIAN=LITTLE"
export CMAKE_OPTIONS="-DENDIAN=NATIVE"

This gives us the flexibility to have machine/application dependent endianness without hardcoding it in the WW3. What do you think? Does it meet your requirements?

ukmo-ccbunney:
> @thesser1 @maryanderson64 @ukmo-ccbunney @mickaelaccensi @sbrus89 Here, we added an option for endianness, so users can define it during the build and select one of BIG, LITTLE and NATIVE options. export CMAKE_OPTIONS="-DENDIAN=BIG" export CMAKE_OPTIONS="-DENDIAN=LITTLE" export CMAKE_OPTIONS="-DENDIAN=NATIVE"
> 
> This gives us the flexibility to have machine/application dependent endianness without hardcoding it in the WW3. What do you think? Does it meet your requirements?

It looks great to me. Nice one @kgerheiser  and @aliabdolali .

sbrus89:
Looks good to me too!

kgerheiser:
Done

aliabdolali:
@kgerheiser thanks for reverting it back. It seems it fixes the problem in mod_def and grid info but caused the same issue we had. 
test with big-endian: The only changes are in ww3_prep/ww3_prnc binary files. The results seem reversed (The data looks the same but is ordered differently). Therefore it changed the results. 
image
Here is the summary of regtests:

[matrixCompFull.txt](https://github.com/NOAA-EMC/WW3/files/8688935/matrixCompFull.txt)
[matrixCompSummary.txt](https://github.com/NOAA-EMC/WW3/files/8688936/matrixCompSummary.txt)
[matrixDiff.txt](https://github.com/NOAA-EMC/WW3/files/8688937/matrixDiff.txt)

kgerheiser:
What are those files and where are they written from?

aliabdolali:
> What are those files and where are they written from?

These files are forcing files for current, ice, wind, mud, density, etc in binary format. You should execute ww3_prep or ww3_prnc to generate them. 

To reproduce it, you can run the develop once and then one with big-endian option from your branch.
` ./bin/run_cmake_test -b slurm -o all -S -T                        -w work_PR3_UQ                 ../model ww3_tp2.8`

kgerheiser:
But do you know where in the code those files are written out?

aliabdolali:
> But do you know where in the code those files are written out?

They should be written in model/src/ww3_prep.F90 and model/src/ww3_prnc.F90
I compared the binary ice.ww3, current.ww3 and wind.ww3, ... with little and big endians, they are identical. so it means something is not right.

kgerheiser:
What CMake option are you running with? Is it set to `BIG`? By making it explicit in the open I don't see how there can be ambiguity. If the files are backwards and we explicitly say `big_endian`/`little_endian` then the opposite is going on currently which also doesn't make any sense because of the hardcoded `-fconvert=`. Somehow there is a mix of files being written out.

aliabdolali:
I agree and that's why I am confused,
I ran once

export CMAKE_OPTIONS="-DENDIAN=BIG"
./bin/run_cmake_test -b slurm -o all -S -T -w work_PR3_UQ_Big ../model ww3_tp2.8

and 

export CMAKE_OPTIONS="-DENDIAN=LITTLE"
./bin/run_cmake_test -b slurm -o all -S -T -w work_PR3_UQ_Little ../model ww3_tp2.8

kgerheiser:
And it shows up in both variants of `BIG`/`LITTLE`? You might expect it to work with one but not the other, but if it's both then it's actually using the opposite being used.

aliabdolali:
> And it shows up in both variants of `BIG`/`LITTLE`? You might expect it to work with one but not the other, but if it's both then it's actually using the opposite being used.

Yes, I was expecting different behaviors. For out_grd.ww3 and out_pnt.ww3 and restart.ww3 files, they are different as expected. But the outputs of ww3_prep and ww3_prnc are the same, and I think this is where we need to fix it.

mickaelaccensi:
> @thesser1 @maryanderson64 @ukmo-ccbunney @mickaelaccensi @sbrus89 Here, we added an option for endianness, so users can define it during the build and select one of BIG, LITTLE and NATIVE options. export CMAKE_OPTIONS="-DENDIAN=BIG" export CMAKE_OPTIONS="-DENDIAN=LITTLE" export CMAKE_OPTIONS="-DENDIAN=NATIVE"
> 
> This gives us the flexibility to have machine/application dependent endianness without hardcoding it in the WW3. What do you think? Does it meet your requirements?

I agree with this solution

aliabdolali:
@kgerheiser I still cannot figure out why big and little are not different with some binaries. 
Here is my question:
Can we just specify the endianness in `model/src/CMakeLists.txt` for endianness, and do not force it in source routines? I mean if we compile it with big or little or native, as CMAKE_OPTIONS="-DENDIAN=big/little/native" only, does it behave differently when it dumps binaries?

Also, I noticed the following a couple of missing endianness options:
model/src/gx_outf.F90:688:              'OPTIONS   big_endian'/                                 &
model/src/gx_outp.F90:503:              'OPTIONS   big_endian'/                                 &

kgerheiser:
@aliabdolali by specify in CMakeLists.txt do you mean using a compiler flag? That's how it's currently done, but the issue is with GNU which applies the `-fconvert=big_endian` only to the main program and not how each individual file was compiled. So, it has to be explicit somewhere in the code.

aliabdolali:
> @aliabdolali by specify in CMakeLists.txt do you mean using a compiler flag? That's how it's currently done, but the issue is with GNU which applies the `-fconvert=big_endian` only to the main program and not how each individual file was compiled. So, it has to be explicit somewhere in the code.
Yes, I thought it should work with compile flag.
ok, I see, and we still need to find why little/big are producing identical binaries.

junwang-noaa:
@aliabdolali May I ask which binary files are the same when you use CMAKE_OPTIONS="-DENDIAN=big/little/native"? Also can you show us which code writes out those files? Thanks.

aliabdolali:
@junwang-noaa 
These files are forcing files for current.ww3, ice.ww3, wind.ww3, mud.ww3, density.ww3, etc in binary format. You should execute ww3_prep or ww3_prnc to generate them (model/src/ww3_prep.F90 and model/src/ww3_prnc.F90 are the programs).

Here is an example:
export CMAKE_OPTIONS="-DENDIAN=BIG"
./bin/run_cmake_test -b slurm -o all -S -T -w work_PR3_UQ_Big ../model ww3_tp2.8
and
export CMAKE_OPTIONS="-DENDIAN=LITTLE"
./bin/run_cmake_test -b slurm -o all -S -T -w work_PR3_UQ_Little ../model ww3_tp2.8

JessicaMeixner-NOAA:
@mickaelaccensi pointed out that he thinks that the opens in https://github.com/NOAA-EMC/WW3/blob/develop/model/src/w3fldsmd.F90#L362-L383 is what is missing for the prep/prnc.

mickaelaccensi:
I think we should add the option CONVERT='BIG_ENDIAN' in the OPEN statements

junwang-noaa:
Thanks, @JessicaMeixner-NOAA @mickaelaccensi. Besides  w3fldsmd.F90, can someone take a look at mod_fileio.f90 too?

It looks to me the following files have open statement with unformatted form for binary file:
gx_outf.F90 
gx_outp.F90
mod_fileio.f90
w3arrymd.F90
w3fldsmd.F90
w3gkemd.F90
w3gridmd.F90
w3iobcmd.F90
w3iogomd.F90
w3iogrmd.F90
w3iopomd.F90
w3iorsmd.F90
w3iosfmd.F90
w3iotrmd.F90
w3smcomd.F90
w3snl4md.F90
w3snl5md.F90
w3src4md.F90
w3srcemd.F90
w3strkmd.F90
wminitmd.F90
ww3_bounc.F90
ww3_bound.F90
ww3_gint.F90
ww3_gspl.F90
ww3_outf.F90
ww3_outp.F90
ww3_prep.F90
ww3_prnc.F90
ww3_shel.F90
ww3_systrk.F90
ww3_trck.F90
ww3_trnc.F90

JessicaMeixner-NOAA:
@junwang-noaa great catch! I agree that mod_fileio needs to be updated for the unformatted open statements.  Similarly to the open statements in w3fldsmd.F90 they need to have if statements for adding the convert statement only if the format variable is unformatted.    

I've looked through all the open statements as well and I think these are the last two files that have missing statements.

kgerheiser:
@JessicaMeixner-NOAA I added it to the open in w3fldsmd.F90 and then reverted it.

https://github.com/NOAA-EMC/WW3/pull/680#issuecomment-1125627048

https://github.com/NOAA-EMC/WW3/pull/680/commits/013482071f56e7e39c1187231a73391b8a298d4c

JessicaMeixner-NOAA:
> @JessicaMeixner-NOAA I added it to the open in w3fldsmd.F90 and then reverted it.
> 
> [#680 (comment)](https://github.com/NOAA-EMC/WW3/pull/680#issuecomment-1125627048)
> 
> [0134820](https://github.com/NOAA-EMC/WW3/commit/013482071f56e7e39c1187231a73391b8a298d4c)

Thanks @kgerheiser I didn't see that from before.  I'm wondering if there should have been two additional convert statements on these open commands: https://github.com/NOAA-EMC/WW3/blob/develop/model/src/w3fldsmd.F90#L375-L382 and that was the issue that was seen before?

ukmo-ccbunney:
> @mickaelaccensi pointed out that he thinks that the opens in https://github.com/NOAA-EMC/WW3/blob/develop/model/src/w3fldsmd.F90#L362-L383 is what is missing for the prep/prnc.

I can confirm that I am seeing the same issues with input fields not having their endian changed.

When I added CONVERT statements to the w3fldsmd.f90 file as mentioned above the issue was resolved (tested with ww3_tp1.6)

For my little endian machine, running with `-DENDIAN="LITTLE"` and `-DENDIAN="NATIVE"` produce identical results.
When using `-DENDIAN="BIG"` the reults are different: I can see with hexdump that the bytes order is reversed, as expected..

aliabdolali:
Thanks Chris, we will check it for a couple of other cases.

junwang-noaa:
@aliabdolali May I ask if the tests you run finished? Is there any issue? Thanks

aliabdolali:
> @aliabdolali May I ask if the tests you run finished? Is there any issue? Thanks

It looks ok from the runs that passed, but do not want to get excited until all tests are done and compared with the base. Stay tuned, it will be done tonight.

aliabdolali:
@kgerheiser @junwang-noaa 
Here are the changes. 
Kyle added two converts in a previous commit, then we reverted it back. The problem is almost solved by adding 4 converts, as shown below:

```
diff model/src/w3fldsmd.F90 Dev/model/src/w3fldsmd.F90
206d205
<       use constants, only: file_endian
363c362
<               OPEN (NDS,FILE=FPRE//FNAME(:I),FORM=FORM,convert=file_endian,ERR=803,       &
---
>               OPEN (NDS,FILE=FPRE//FNAME(:I),FORM=FORM,ERR=803,       &
369c368
<               OPEN (NDS,FILE=FNAME(:I),FORM=FORM,convert=file_endian,ERR=803,IOSTAT=IERR)
---
>               OPEN (NDS,FILE=FNAME(:I),FORM=FORM,ERR=803,IOSTAT=IERR)
376c375
<               OPEN (NDS,FILE=FPRE//FNAME(:I),FORM=FORM,convert=file_endian,               &
---
>               OPEN (NDS,FILE=FPRE//FNAME(:I),FORM=FORM,               &
382c381
<               OPEN (NDS,FILE=FNAME(:I),FORM=FORM, convert=file_endian,                     &
---
>               OPEN (NDS,FILE=FNAME(:I),FORM=FORM,                     &
```

Kyle, could you do these changes, and make your branch up to date with develop? I tried to make a PR to your branch with no success. Your fork was not shown on my repo.

Now, I ran the entire WW3 regression test with *endian_big* and compared it with the base:
everything is expected except 
```
ww3_ts1/./work_ST4_TSA                     (9 files differ)
ww3_ts1/./work_NL5                     (10 files differ)
ww3_ts1/./work_ST4_WRT                     (1 files differ)
```

[matrixCompFull.txt](https://github.com/NOAA-EMC/WW3/files/8723076/matrixCompFull.txt)
[matrixCompSummary.txt](https://github.com/NOAA-EMC/WW3/files/8723077/matrixCompSummary.txt)
[matrixDiff.txt](https://github.com/NOAA-EMC/WW3/files/8723078/matrixDiff.txt)


I then ran the entire regtests with **endian_little** and compared with *endian_big*
the binaries are different, but netcdf and grib2 files are identical as expected.

[matrixCompFull.txt](https://github.com/NOAA-EMC/WW3/files/8723085/matrixCompFull.txt)
[matrixCompSummary.txt](https://github.com/NOAA-EMC/WW3/files/8723086/matrixCompSummary.txt)
[matrixDiff.txt](https://github.com/NOAA-EMC/WW3/files/8723087/matrixDiff.txt)


@ukmo-ccbunney @mickaelaccensi, any idea why `ww3_ts1` cases are different?

junwang-noaa:
@aliabdolali Thanks for running the test. From the changes you listed, it seems to me that those files (without convert=file_endian) will be written out with machine endianness, which is a little endian on the current noaa r&d platforms. Can you check if those files are little endian in your baseline? It's still not clear to me why you need to revert those changes.

JessicaMeixner-NOAA:
@aliabdolali I agree with @junwang-noaa that there's likely a missing file that is not getting converted.  However for the ts1 cases that are different the mod_defs are also different, which that should have been taken care of so that's a little odd.  Can you share the complete paths of your regression tests so I can take a look at the ww3_grid.out files to see if that gives us any clues?

aliabdolali:
@JessicaMeixner-NOAA the problem is the mod_def is written differently, so I agree that something is missing in the mod_def and led to changes in results. Nonlinear interactions (GMD vs TDA).
As you can see here, a different physics is selected with big_endian compared to the base 

```
68c68
<   Nonlinear interactions (GMD) (default values) :
---
>   Nonlinear interactions (TSA) (default values) :
70,74c70,72
<        Powers in scaling functions :    0.00  -3.50
<        Nondimension filter depths  :    0.20   5.00
<        Number of quad. definitions :    1
<                 0.250   0.000  -1.0  0.1000E+08  0.0000E+00
<
---
>    Source term computation (1=TSA,0=FBI) :  1
>    Alternate loops (1=no,2=yes)          :  2
>    (To speed up computation)
155,156c153
<   &SNL3 NQDEF =  1, MSC =  0.00,  NSC = -3.50,  KDFD =  0.20,  KDFS =  5.00 /
<   &ANL3 QPARMS = 0.250, 0.000,  -1.0, 0.1000E+08, 0.0000E+00 /
---
>   &SNL4 ITSA = 1, IALT = 2**
**
```



The info of run are in the matrixComp files:
big endian runs:
/scratch1/NCEPDEV/stmp2/Ali.Abdolali/Source/May19endian/regtests/
little endian runs:
/scratch1/NCEPDEV/stmp2/Ali.Abdolali/Source/May19endianLittle/regtests/
base (develop)
/scratch1/NCEPDEV/stmp2/Ali.Abdolali/Source/May18Dox1/regtests/

aliabdolali:
something in mod_def is written wrong, from TSA (dev) to DIA (big_endian/little_endian).

image

JessicaMeixner-NOAA:
I'm currently running two sets of regtests with develop to see if we see any issues there.  In addition I'm running just some ts1 tests to look at how things are being set-up as it does not appear to be set-up as expected based on the mod_def differences.

aliabdolali:
> I'm currently running two sets of regtests with develop to see if we see any issues there. In addition I'm running just some ts1 tests to look at how things are being set-up as it does not appear to be set-up as expected based on the mod_def differences.

@JessicaMeixner-NOAA , I found the problem, the switch is different, NL3 (DIA) and NL4 (TSA) are changed. I just need to check when it changed. Stay tuned.

JessicaMeixner-NOAA:
Interesting... This branch has not have the develop branch merged into it recently.  Perhaps that is an issue?

aliabdolali:
> Interesting... This branch has not have the develop branch merged into it recently. Perhaps that is an issue?

I don't think that's the case.
The switches are the same between develop and endian branch for ww3_ts1/./work_ST4_TSA, ww3_ts1/./work_NL5 and ww3_ts1/./work_ST4_WRT. 
However, if you look at mod_def.ww3 binaries, the endian branch is the wrong one and develop is picking the right physics. 

Also, the switch is shown in the metadata of netcdf output files, for endian branch it is wrong and correct for develop.

JessicaMeixner-NOAA:
So I have made some updates which can be found here: https://github.com/jessicameixner-noaa/ww3/tree/explicit-big-endian 

and then just running the ts1 tests Ali noted with differences, I now only have the diff: 
`ww3_ts1/./work_ST4_WRT                     (1 files differ)
`which is: 
`     quad99999.bqf differ (binary)
`

JessicaMeixner-NOAA:
@kgerheiser do you just want to merge my updates from https://github.com/jessicameixner-noaa/ww3/tree/explicit-big-endian or do you want me to make a PR to your branch?  I'm ready for the last round of testing with this branch.  If you'd like you can wait until we finish the full set of regression tests, but this resolves the issues from the regression test results Ali mentioned earlier today.

kgerheiser:
Yeah, I can just merge it.

JessicaMeixner-NOAA:
In the meantime, I've started a full set of WW3 regression tests, will start a set of ufs-weather-model develop branch + this ww3 branch, and then test @DeniseWorthen ufs-weather-model branch + ww3 branch with this merged in for GNU. Hopefully we will not encounter any surprises.

aliabdolali:
@kgerheiser @JessicaMeixner-NOAA 
thanks both of you for synicing the develop and fixing the remained issues. I just submitted the matrix.

DeniseWorthen:
I've been able to compile and run S2SW with GNU on Cheyenne for the mesh cap PR after merging @JessicaMeixner-NOAA branch.

aliabdolali:
The entire WW3 regression passed successfully with pre-known non-identical cases:
```

mww3_test_03/./work_PR2_UNO_MPI_d2                     (15 files differ)
mww3_test_03/./work_PR3_UQ_MPI_d2_c                     (16 files differ)
mww3_test_03/./work_PR1_MPI_d2                     (16 files differ)
mww3_test_03/./work_PR3_UNO_MPI_d2_c                     (17 files differ)
mww3_test_03/./work_PR3_UNO_MPI_d2                     (15 files differ)
mww3_test_03/./work_PR3_UQ_MPI_d2                     (15 files differ)
ww3_tnc1/./work                     (16 files differ)
ww3_ufs1.3/./work_a                     (1 files differ)


```
[matrixCompFull.txt](https://github.com/NOAA-EMC/WW3/files/8735717/matrixCompFull.txt)
[matrixCompSummary.txt](https://github.com/NOAA-EMC/WW3/files/8735718/matrixCompSummary.txt)
[matrixDiff.txt](https://github.com/NOAA-EMC/WW3/files/8735719/matrixDiff.txt)

aliabdolali (approved):

kgerheiser:
Thank you @aliabdolali  and @JessicaMeixner-NOAA for helping with this PR.

JessicaMeixner-NOAA:
@aliabdolali  is 

`ww3_tnc1/./work                     (16 files differ)`

an expected or unexpected change?  I didn't know that was expected to be different?

aliabdolali:
@JessicaMeixner-NOAA good catch, I'll check, from what I remember, it was affected by namelist commit (tn1 is one of -N cases in matrix.base).

JessicaMeixner-NOAA:
> @JessicaMeixner-NOAA good catch, I'll check, from what I remember, it was affected by namelist commit (tn1 is one of -N cases in matrix.base).

Hopefully we can fix this ASAP.  This shouldn't have been merged as is and should hopefuly be fixed before this is merged to def/ufs-weather-model.  @aliabdolali can you help with debugging?  FYI @AvichalMehra-NOAA

aliabdolali:
@JessicaMeixner-NOAA you ran this PR yesterday, did you see the same in your comparison?

JessicaMeixner-NOAA:
> @JessicaMeixner-NOAA you ran this PR yesterday, did you see the same in your comparison?

You had merged the PR, therefore I deleted my directory.  I should have a comparison of the two develop branches in the next hour or so.  In the meantime, its a priority to continue to dig into this issue and see if it can be fixed so the fix can be included in the ufs-weather-model PR you have.  When is the commit date for that?

aliabdolali:
> > @JessicaMeixner-NOAA you ran this PR yesterday, did you see the same in your comparison?
> 
> You had merged the PR, therefore I deleted my directory. I should have a comparison of the two develop branches in the next hour or so. In the meantime, its a priority to continue to dig into this issue and see if it can be fixed so the fix can be included in the ufs-weather-model PR you have. When is the commit date for that?

It should be determined by @junwang-noaa
I'd say we still can continue, and merge dev/ufs-weather-model. This test does not affect that side and I already tested ufs-weather-model. In the meantime, I will take a look based on my priorities.

aliabdolali:
@JessicaMeixner-NOAA FYI: two separate runs from the develop for tnc1 are the same, need to check with commits before endian, namelist, scal.

ukmo-ccbunney:
Just a note to say that this PR has broken the "gmake" build of the model (the ENDIANNESS variable is not expanded in constants.ftn).

I don't know if we care about this considering we will be retiring gmake builds soon.... (forced early retirement!?)

aliabdolali:
@ukmo-ccbunney good point Chris, we are close to June 1st when we plan to start retiring gnu make. Do you have a fix for it? so I can put it as a priority and merge it.

ukmo-ccbunney:
I think I have a fix. I'll raise a PR.

PR #743, opened 2022-07-25
Feature/gint ice mask

### LINKED ISSUE #316: gint ice mask
In ww3_gint, forcings parameters (winds, ice, etc) are masked where ice concentration is greater than CICEN. 

This is not the case for output directly out of ww3_shel, where we get wind and ice output even for points where ice concentration is greater than CICEN. I would like to see this behaviour in gint.

This is with IC0.

benoitp-cmc:
# Pull Request Summary

In gint, do not automatically set points as inactive if sea ice concentration > CICEN. Add PNR.

## Description

In ww3_gint, forcings parameters (winds, ice, etc) are masked where ice concentration is greater than CICEN.

This is not the case for output directly out of ww3_shel, where we get wind and ice output even for points where ice concentration is greater than CICEN. This PR makes gint behave the same way as shel.

This does not change behaviour for wave fields which continue to be masked where ice concentration is greater than CICEN.

Add PNR.

Please also include the following information: 
* Add any suggestions for a reviewer @aliabdolali @JessicaMeixner-NOAA @mickaelaccensi 
* Mention any labels that should be added:  _enhancement_
* Are answer changes expected from this PR? _Regression test_

### Issue(s) addressed

- fixes #316

### Commit Message

In gint: do not automatically set points as inactive if sea ice concentration > CICEN. Add PNR.

### Check list  

- [x] Branch is up to date with the authoritative repository (NOAA-EMC) develop branch. 
- [x] Checked the [checklist for a developer submitting to develop](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-a-developer-submitting-to-develop). 
 
### Testing

* Have the matrix regression tests been run (if yes, please note HPC and compiler)? No

### DIFF
--- a/model/src/ww3_gint.F90
+++ b/model/src/ww3_gint.F90
@@ -26,7 +26,7 @@
 !/                  | WAVEWATCH-III           NOAA/NCEP |
 !/                  |             A. Chawla             |SX
 !/                  |                        FORTRAN 90 |
-!/                  | Last update :         02-Jun-2021 |
+!/                  | Last update :         25-Jul-2022 |
 !/                  +-----------------------------------+
 !/
 !/    15-Mar-2007 : Origination.                        ( version 3.13 )
@@ -42,8 +42,7 @@
 !/    26-Jan-2021 : Added TP field (derived from FP)    ( version 7.12 )
 !/    22-Mar-2021 : New coupling fields output          ( version 7.13 )
 !/    02-Jun-2021 : Bug fix (*SUMGRD; Q. Liu)           ( version 7.13 )
-!/    18-Feb-2021 : Modify to keep ice points active    ( version 7.12 )
-!/                  unless masked (B. Pouliot, CMC)
+!/    25-Jul-2022 : Keep ice points active (B. Pouliot) ( version 7.14 )
 !/
 !   1. Purpose : 
 !
@@ -956,6 +955,7 @@
 !/    30-Apr-2014 : Add group 3                         ( version 5.00 )
 !/    27-Aug-2015 : ice thick. and floe added as output ( version 5.10 )
 !/    22-Mar-2021 : New coupling fields output          ( version 7.13 )
+!/    25-Jul-2022 : Add PNR. ACTIVE label for ice       ( version 7.14 )
 !/
 !   1. Purpose :
 !
@@ -1036,7 +1036,7 @@
         REAL          :: PHSAUX(0:NOSWLL_MIN), PTPAUX(0:NOSWLL_MIN),   &
                          PLPAUX(0:NOSWLL_MIN), PSIAUX(0:NOSWLL_MIN),   &
                          PWSAUX(0:NOSWLL_MIN), PDIRAUX1(0:NOSWLL_MIN), &
-                         PWSTAUX, PDIRAUX2(0:NOSWLL_MIN),              &
+                         PWSTAUX, PNRAUX, PDIRAUX2(0:NOSWLL_MIN),      &
                          PTHP0AUX1(0:NOSWLL_MIN),                      &
                          PTHP0AUX2(0:NOSWLL_MIN),                      &
                          PQPAUX(0:NOSWLL_MIN), PPEAUX(0:NOSWLL_MIN),   &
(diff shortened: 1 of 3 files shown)


JessicaMeixner-NOAA:
@benoitp-cmc I have run the full set of regression tests and the changes are likely expected: 
[matrixCompFull.txt](https://github.com/NOAA-EMC/WW3/files/9199497/matrixCompFull.txt)
[matrixCompSummary.txt](https://github.com/NOAA-EMC/WW3/files/9199499/matrixCompSummary.txt)
 as they look to be files that just have been processed by ww3_gint, however I'm not seeing what I would expect to see when I look at the output itself.  I would not have expected value changes in HS for example, and I don't see huge changes in say ice concentration or other input fields in the combined files.  I don't know if I'm misunderstanding the desired change or if there's something going wrong.   So I was hoping maybe you could provide an example or point to a regression test that would help illustrate the behavior you are expecting so that I can confirm that's what I'm also seeing.

benoitp-cmc:
There should be no changes in Hs or any of the wave parameters. It really should only affect group 1. I'll run more tests on my side.

benoitp-cmc:
@JessicaMeixner-NOAA I now see the output of `ww3_outf` and `ww3_ounf` is always masked when ice concentration is greater than CICEN.

At CMC, we use our own internal file format. It's generated with a program that diverged from `ww3_outf` circa version 3.14 or 4.16. With our internal file format I don't mask behind ICEC > CICEN. Without this PR, `ww3_gint` overrides that by imposing the ice mask. 

For pure wave modelling application it's not a big deal. However, our users like having the context of winds and ice concentration over the ice.

I'll try to dig a bit more to see why ounf and outf mask things like this.

JessicaMeixner-NOAA:
@benoitp-cmc I can see the benefit in having input fields not be masked out when  ICEC > CICEN.   It'll be great to get your diverged updates back into the main repo so that you don't have this issue again.  Let me know if you end up needing an extra set of eyes to look at the other files for updates, hopefully if anyone objects to these changes they'll comment in the issue.  We could make an extra comment to say that these changes will be extended to other programs?

benoitp-cmc:
@JessicaMeixner-NOAA `ww3_ounf` and `ITYPE=[23]` of `ww3_outf` mask wherever `MAPSTA(IY,IX) .LE. 0` (wet with ice: `MAPSTA=-1`). If we replace `.LE.` with `.EQ.` in those conditionals, we get the forcings in the output. 

That being said, I don't think the check on MAPSTA is useful:
* It's always followed by a check like `XX(IX,IY) .EQ. UNDEF`
* The MAPSTA check is not used for the SMC grid (@ukmo-ccbunney technical reason but ounf works as expected?) 
* `ww3_grib` does not use MAPSTA, it only does the UNDEF check to define the BITMAP.

If the check on MAPSTA can be dropped, we could probably remove many duplicated lines in `ww3_ounf` between the SMC and non-SMC code.

ukmo-ccbunney:
> * The MAPSTA check is not used for the SMC grid (@ukmo-ccbunney technical reason but ounf works as expected?)

@benoitp-cmc That's right - MAPSTA is not used when in ww3_ounf when using the SMC grid. This is because the regular grid output option is in fact an interpolated output (generating regular grid cells from irregular SMC cells), so MAPSTA no longer makes any sense.

JessicaMeixner-NOAA:
> 

I need to go back and carefully review the code, but this makes sense to me to remove the MAPSTA check.  I'm wondering if someone was trying to reduce file output size? Plus removing duplicative lines always sounds nice.

benoitp-cmc:
In v1.18, `ww3_outf` only had the MAPSTA check. In v3.14 it had both the MAPSTA and the UNDEF check. The UNDEF check probably got added to deal with partionned fields. It's easier to leave a maybe redundant check than risk breaking things.

I've done the cleanup I mentionned, I'll push it later once I've run the regtests and looked at the results.

JessicaMeixner-NOAA:
Thanks @benoitp-cmc !

JessicaMeixner-NOAA:
@benoitp-cmc let me know when you're ready for me to test things again on this end.

benoitp-cmc:
@JessicaMeixner-NOAA I am tempted to close it and to revisit later. If we change outf and ounf to output the ice covered points, it changes a lot of regression tests. I'm not sure all those changes are good. 

For the gint side of things, I think I could have been smarter. And it probably could be two pull requests.

JessicaMeixner-NOAA:
@benoitp-cmc it's completely up to you, just let me know how I can test and help on my end.

benoitp-cmc:
I'll close this and do it better later, after the ifdef effort. Thanks.

Issue #678, opened 2022-04-20
Create ww3_shel.inp file and remove set_shel_inp 


JessicaMeixner-NOAA:
**Describe the bug**
@MatthewMasarik-NOAA  will provide a ww3_shel.inp file that CESM can use and remove the need for the set_shel_inp routine.  This will allow users of the CESM coupled system to have additional flexibility without recompiling the code. 

**Additional context**
This is part of merging https://github.com/NOAA-EMC/WW3/pull/641 to the develop branch

mvertens:
@JessicaMeixner-NOAA - is this needed for PR #641 - or can we put an #ifdef W3_CESMCOUPLED in for now. @alperaltuntas - I think that once a new ww3_shel.inp file is created it would be easy to read it in if its ascii. But we need to have the mechanism to recreate it on our end flexibly for the different use cases. How hard do you think that is going to be?

JessicaMeixner-NOAA:
@mvertens this is not needed for PR #641 which is going into the dev/ufs-weather-model branch.  This will need to be addressed before the contents of PR #641 can be merged into the develop branch though.

mvertens:
@JessicaMeixner-NOAA - thanks for the clarification.

MatthewMasarik-NOAA:
@JessicaMeixner-NOAA , it doesn't look like I have permissions to add myself as an assignee to this issue.  Would you be able to add me?

MatthewMasarik-NOAA:
[ww3_shel_mesh.inp.txt](https://github.com/NOAA-EMC/WW3/files/9055146/ww3_shel_mesh.inp.txt)

MatthewMasarik-NOAA:
Corrected (blank lines removed).
[ww3_shel_mesh.inp.txt](https://github.com/NOAA-EMC/WW3/files/9055254/ww3_shel_mesh.inp.txt)

Issue #679, opened 2022-04-20
Separate the reading of ww3_shel.inp 


JessicaMeixner-NOAA:
**Is your feature request related to a problem? Please describe.**
Create a separate routine to read ww3_shel.inp so that both ww3_shel program and the mesh based cap can use the same routine for reading the ww3_shel.inp file (and ww3_shel.nml).  

**Describe the solution you'd like**
New subroutine to read ww3_shel.inp to reduce repeated code, which will help with code maintenance. 

**Describe alternatives you've considered**
All initialization needs for ww3_shel.inp could be moved into a subroutine. 

**Additional context**
This is part of merging https://github.com/NOAA-EMC/WW3/pull/641 to the develop branch

arunchawla-NOAA:
There is already a function to read ww_shel.inp data in wave_shel_inp.F90, is the request to use this function in ww3_shel.F90?

JessicaMeixner-NOAA:
@arunchawla-NOAA yes, this way we don't have duplicative code and if code needs updated its done once for both places.

MatthewMasarik-NOAA:
@arunchawla-NOAA, this has been my task.  I've started it but was pulled to higher priority items.  It will take a small dedicated  chunk of time that I can move through the duplicative regions and refactor the duplicated code out.

Issue #676, opened 2022-04-20
Remove ifdefs around IC4METHOD = 8 


JessicaMeixner-NOAA:
**Is your feature request related to a problem? Please describe.**
Remove the ifdefs around IC4METHOD = 8 so that this option can be used for non CESMCOUPLED cases. 

**Additional context**
This is part of merging https://github.com/NOAA-EMC/WW3/pull/641 to the develop branch, but should be submitted to the develop branch as it's own self-contained PR.

JessicaMeixner-NOAA:
@JessicaMeixner-NOAA will check if this is still an issue and provide more info.  There might be a update for this in a branch from @DeniseWorthen

DeniseWorthen:
My working branch is [feature/wave_ice_only](https://github.com/DeniseWorthen/WW3/tree/feature/wave_ice_only).

MatthewMasarik-NOAA:
Thank you @DeniseWorthen

JessicaMeixner-NOAA:
This is taken care of @DeniseWorthen branch see: https://github.com/DeniseWorthen/WW3/blob/feature/wave_ice_only/model/src/w3sic4md.F90#L523 where it's commented out.  Thanks @DeniseWorthen !

JessicaMeixner-NOAA:
Completed in PR #780

Issue #674, opened 2022-04-20
Allow for user-configured naming conventions of restart files 


JessicaMeixner-NOAA:
**Is your feature request related to a problem? Please describe.**
CESM and other systems have requirements for specific naming conventions of restart files.   

**Describe the solution you'd like**
Configure the name of restart files via optional namelist/input options. 

**Additional context**
Need to know what CESM (@alperaltuntas) and other systems requirements are. 

Related to https://github.com/NOAA-EMC/WW3/issues/219

alperaltuntas:
Thanks @JessicaMeixner-NOAA for posting this! CESM naming convention of restart file names is as follows:

`$CASE.$scomp.r.$date[$ending] `

where, 
`$CASE` is the experiment name,
`$scomp` is the component name, e.g., `cam`, `mom`, `ww3`, etc.
`$date` is the model date string which is based on a yyyy-mm-dd-sssss convention
`$ending` is the file extension, which is ideally .nc 

----

An example ww3 restart file name from a CESM case that I recently ran:

`g.e23.GMOM_JRA.TL319_t061_wt061.control.001.ww3.r.0060-01-01-00000`

where `$CASE` is `g.e23.GMOM_JRA.TL319_t061_wt061.control.001`,
`$scomp` is `ww3`,
`$date` is  `0060-01-01-00000`
and `$ending` is empty since ww3 files are binary files as opposed to netcdf file (which is another feature we would love to have.)

CESM has a unique naming convention for case names as well. As in the above example, we tend to have quite long $CASE names: `g.e23.GMOM_JRA.TL319_t061_wt061.control.001`. The full description of CESM naming convention is described here: https://www.cesm.ucar.edu/models/cesm2/naming_conventions.html 

Within other CESM components, we achieve this naming convention in the respective model drivers (caps) where we can easily obtain the case name and date strings from the CESM driver at runtime. For instance, in MOM6 nuopc cap, we set the restart filename as follows:

```
 write(restartname,'(A,".mom6.r.",I4.4,"-",I2.2,"-",I2.2,"-",I5.5)') &
             trim(casename), year, month, day, seconds
```

where `casename`, `year`, `month`, `day`, and `seconds` are obtained via `NUOPC_CompAttributeGet` and `ESMF_TimeGet` calls.

----

[...]

PR #780, opened 2022-09-08
Mesh cap update 

DeniseWorthen:
# Pull Request Summary

Update mesh cap with changes needed for eventual merge to develop.

## Description

- Allows users of the mesh cap to set runtime options for custom restart and history file names and to create netcdf output for gridded fields at run time. For UWM, these can be set in the nems.configure using the settings

```
  gridded_netcdfout = true
  user_sets_histname = true
  user_sets_restname = true
```

History file names will then be ``casename.ww3.hi.YYYY-MM-DD-SSSS.nc`` and ``casename.ww3.r.YYYY-MM-DD-SSSS`` where ``casename`` is set from nems.configure for each application (e.g ``ufs.hafs``)

- Replaces previous ``set_shel_inp`` and ``read_shel_inp`` routines with the applicable section from ``ww3_shel.F90`` to read either a inp or nml file. When used in UFS, the only required fields in the nml file are the list of forcing fields, the desired output fields, the name of the points file (if desired) and the stride for history and restart files. For example,

```
&input_nml
  input%forcing%winds = 'C'
  input%forcing%currents  = 'F'
  input%forcing%ice_conc  = 'F'
/
&output_type_nml
  type%field%list     = 'WND HS FP DP PHS PTP PDIR'
  type%point%file     = 'points.list'
/

&output_date_nml
  date%field%outffile  = '1'
  date%field%stride    = '3600'
  date%point%stride    = '3600'
  date%restart2%stride = '3600'
/
```

- Removes ``wav_tauice`` fields which are unused when coupling WW3 to CICE6.

### Issue(s) addressed
- https://github.com/NOAA-EMC/WW3/issues/679
- https://github.com/NOAA-EMC/WW3/issues/674
- https://github.com/NOAA-EMC/WW3/issues/673
- https://github.com/NOAA-EMC/WW3/issues/676

### Commit Message

Update mesh cap for custom user filenames, netcdf gridded output and logical control of some ifdef options.

co-authors : @mvertens 

### Testing

All UFS-WM baselines were run using hera.intel and hera.gnu and were B4B.


model/src/constants.F90
```diff
@@ -20,7 +20,7 @@ MODULE CONSTANTS
 !/
 !/    Copyright 2009-2012 National Weather Service (NWS),
 !/       National Oceanic and Atmospheric Administration.  All rights
-!/       reserved.  WAVEWATCH III is a trademark of the NWS. 
+!/       reserved.  WAVEWATCH III is a trademark of the NWS.
```
JessicaMeixner-NOAA: This entire file is just white space changes.  Should this go ahead and be reverted?

model/src/w3fldsmd.F90
```diff
@@ -25,10 +25,10 @@ MODULE W3FLDSMD
 !/    06-Dec-2010 : Change from GLOBAL (logical) to ICLO (integer) to
 !/                  specify index closure for a grid.   ( version 3.14 )
 !/                  (T. J. Campbell, NRL)
-!/    30-Oct-2012 : Implement tidal analysis            ( version 4.08 ) 
+!/    30-Oct-2012 : Implement tidal analysis            ( version 4.08 )
```
JessicaMeixner-NOAA: This entire file is just white space changes.  Should this file be reverted to it's original?

model/src/w3servmd.F90
```diff
@@ -10,9 +10,9 @@ MODULE W3SERVMD
 !/                  +-----------------------------------+
 !/
 !/    For update log see individual subroutines.
-!/    12-Jun-2012 : Add /RTD option or rotated grid option. 
+!/    12-Jun-2012 : Add /RTD option or rotated grid option.
```
JessicaMeixner-NOAA: Most changes are white space or style and are to me seem unrelated to the PR.   I'm wondering if it's worth keeping these changes in the PR or if it would be better to revert this file to its original state which would help keep this branch and develop in sync.

model/src/w3adatmd.F90
```diff
@@ -340,6 +340,9 @@ MODULE W3ADATMD
 !  7. Source code :
 !
 !/ ------------------------------------------------------------------- /
+      ! module default
```
JessicaMeixner-NOAA: Most changes are white space or style and to me seem unrelated to the PR.   I'm wondering if it's worth keeping these changes in the PR or if it would be better to revert this file to its original state which would help keep this branch and develop in sync.

model/src/w3gdatmd.F90
```diff
@@ -46,28 +46,28 @@ MODULE W3GDATMD
 !/    30-Oct-2009 : Implement curvilinear grid type.    ( version 3.14 )
 !/                  (W. E. Rogers & T. J. Campbell, NRL)
 !/    29-Oct-2010 : Implement unstructured grids        ( version 3.14.1 )
-!/                  (A. Roland and F. Ardhuin) 
+!/                  (A. Roland and F. Ardhuin)
```
JessicaMeixner-NOAA: Most changes are white space or style and to me seem unrelated to the PR.   I'm wondering if it's worth keeping these changes in the PR or if it would be better to revert this file to its original state which would help keep this branch and develop in sync.

model/src/w3wavemd.F90
```diff
@@ -697,24 +697,20 @@ SUBROUTINE W3WAVE ( IMOD, ODAT, TEND, STAMP, NO_OUT &
...
-#ifdef W3_SBS
-      NDSOFLG = 99
-#endif
-#ifdef W3_MPI
       SBSED = .FALSE.
-#endif
-#ifdef W3_SBS
-      SBSED = .TRUE.
-#endif
+      if (w3_sbs_flag) then
```
JessicaMeixner-NOAA, line 710: Can this be kept as #ifdef W3_SBS and keep changing any #if-defs to a seperate PR?

model/src/wav_shr_flags.F90
```diff
@@ -0,0 +1,1201 @@
+!> @file wav_shr_flags
```
JessicaMeixner-NOAA, line 1: This seems related to if-def removals and not this PR, but perhaps I've missed something?  Is this file necessary for this PR?

JessicaMeixner-NOAA (changes requested):
The standalone WW3 regression tests should be finishing soon.  These tests will help ensure there are no unintended impacts in the updates in w3wave (which visually all look good so far).  

For issue  #679 @MatthewMasarik-NOAA and I will work to extend the work here to remove duplicated code and have ww3_shel exes use the same separated subroutine for reading ww3_shel.inp. This issue will remain open after this PR. 

For  issue #674 I will be working as I mentioned the other day to extend this work so that the restart file writing will be variables read in from WW3 input files.   This will either change the reading form the nems.configure or allow for the nems.configure to over-ride this.  This issue will remain open after this PR. 

For issue issue #673  Work will be ongoing to extend this capability to reach the desired capabilities mentioned in the issues. 

I do believe this PR will close issue #678  which was not mentioned in the PR. 

I will report on the standalone WW3 regression tests as soon as the script has finished running, but wanted to provide this update now.  Apologies for the delay in review, we'll be working to reduce the review time for PRs in the future and thank you so much for the work you have contributed here!

model/src/constants.F90
DeniseWorthen: This is stripping of trailing white space. Trailing whitespace should not be committed.

model/src/w3fldsmd.F90
DeniseWorthen: Again, this is stripping of trailing whitespace....same with the following white space changes.

model/src/w3wavemd.F90
DeniseWorthen: The issue here is that the sbs if-def is one of those "branching ifdefs" which are very difficult to code around w/o nesting additional ifdefs, which we wanted to avoid.

model/src/wav_shr_flags.F90
DeniseWorthen: We don't need all the flags, but it was simply easier to keep the original list together.

model/src/w3adatmd.F90
DeniseWorthen: implicit none should always be at the module level. Reverting this to the previous would mean reverting to a poor coding practice.

model/src/wav_shr_flags.F90
JessicaMeixner-NOAA: If we backed out the W3_SBS change, can all of this be backed out?

model/src/w3wavemd.F90
JessicaMeixner-NOAA: It'd be great to back this out if possible to keep different development separate.

model/src/w3adatmd.F90
JessicaMeixner-NOAA: From a code management perspective, it's preferred to keep the divergence from the develop branch to a minimum and keep the PR focused as possible.

model/src/constants.F90
JessicaMeixner-NOAA: The preference is to take care of trailing white spaces in it's own PR to the develop branch instead of being combined here.  While github does have a great feature to ignore white spaces in a review, it's still preferred to handle this separately.

JessicaMeixner-NOAA:
The WW3 standalone tests performed as expected.

model/src/w3wavemd.F90
DeniseWorthen: No, it is not really possible to back out the change to this branching ifdef. I believe it would work to remove the wav_shr_flag file and put in a local definition of this flag.

JessicaMeixner-NOAA: I think that sounds like a great path forward!

DeniseWorthen: It turns out I cannot remove wav_shr_flags completely since it is also used by the mesh cap code. I removed it from w3wavemd.F90 and added locally defined flags.

JessicaMeixner-NOAA: Thanks @DeniseWorthen

JessicaMeixner-NOAA:
@DeniseWorthen I really appreciate your updates to this branch.  I'll be re-running the standalone WW3 regression tests but expect no issues there.

DeniseWorthen:
@JessicaMeixner-NOAA @MatthewMasarik-NOAA I had fixed the CI previously for the ufs branch but it is now failing. So I checked out develop and pushed a copy to my fork and it also fails. I don't know why---can one of you figure it out? The branch I created and tested is [testci](https://github.com/DeniseWorthen/WW3/tree/testci). You can see the workflow error in the Intel Linux Build / setup (push)

```
Note that versions of Spack older than 0.19 may not be able to use the updated configuration.
==> Error: list index out of range
Error: Process completed with exit code 3.
```

JessicaMeixner-NOAA:
@DeniseWorthen yes, we can look at the ci for this branch and develop.  Realistically, we will not be able to get it fixed before this branch's turn in the ufs-queue.

JessicaMeixner-NOAA:
Results of standalone regression tests were as expected.

JessicaMeixner-NOAA (approved):
Updates were made based on review and ufs-waether-model regression tests passed (https://github.com/ufs-community/ufs-weather-model/pull/1366) in addition to ww3 standalone.  

Thanks @DeniseWorthen !

PR #852, opened 2022-11-16
Typo correction + moving of code.

MathieuDutSik:
# Pull Request Summary
 
Correct a typo in the text and moved some code in a better position.
The logic is following: SNL1 does not need to have VSNL and others initialized
since it sets up values directly. Also, in case of NL0, the assignation to zero is
best done near the call to SNL1, SNL2 and others.

## Description

No test affected.

### Issue(s) addressed

No issues were raised.

### Check list  

### Testing

### DIFF
--- a/model/src/w3snl1md.F90
+++ b/model/src/w3snl1md.F90
@@ -90,7 +90,7 @@ CONTAINS
     !     frequency grid is invariant, the nonlinear interactions are
     !     calculated for the frequency spectrum, as in WAM. This requires
     !     only a single set of interpolation data which can be applied
-    !     throughout the spatial domain. For deep water this is idenitical
+    !     throughout the spatial domain. For deep water this is identical
     !     to a direct application to the wavenumber spectrum, for shallow
     !     water it is not. As the shallow water correction is nothing but
     !     a crude approximation, the choice between spectra is expected to
--- a/model/src/w3srcemd.F90
+++ b/model/src/w3srcemd.F90
@@ -869,14 +869,6 @@ CONTAINS
     VDIN = 0.
 #endif
 
-#ifdef W3_NL0
-    VSNL = 0.
-    VDNL = 0.
-#endif
-#ifdef W3_NL1
-    VSNL = 0.
-    VDNL = 0.
-#endif
 #ifdef W3_TR1
     VSTR = 0.
     VDTR = 0.
@@ -1254,6 +1246,10 @@ CONTAINS
       !
       ! 2.b Nonlinear interactions.
       !
+#ifdef W3_NL0
+      VSNL = 0.
+      VDNL = 0.
+#endif
 #ifdef W3_NL1
       CALL W3SNL1 ( SPEC, CG1, WNMEAN*DEPTH,        VSNL, VDNL )
 #endif


MatthewMasarik-NOAA:
@MathieuDutSik thank you for this PR.  We will start the review of it.  We'll want to check when/who put the initializations in, and make sure taking them out doesn't have any adverse affects.  From a defensive programming standpoint I generally lean towards putting more initializations in vs. taking them out, we can consider however if it makes sense.

MathieuDutSik:
> @MathieuDutSik thank you for this PR. We will start the review of it. We'll want to check when/who put the initializations in, and make sure taking them out doesn't have any adverse affects. From a defensive programming standpoint I generally lean towards putting more initializations in vs. taking them out, we can consider however if it makes sense.

There are two issues here:
* If a function does `S = 0`. and then `S(ISP) = ....` for all ISP then the first assignation is not useful.
* If we assign `S = 0` then this should be done nearest to the relevant code so as to diminish the number of dependencies.

MatthewMasarik-NOAA:
> There are two issues here:
> 
>     * If a function does `S = 0`. and then `S(ISP) = ....` for all ISP then the first assignation is not useful.
> 
>     * If we assign `S = 0` then this should be done nearest to the relevant code so as to diminish the number of dependencies.

Hi @MathieuDutSik,  I've kicked this around, and I'm not comfortable moving/removing existing initializations.  Putting in initializations during development is not the same thing as removing them after the fact, especially with a large, community code base.  I don't see any gain in this instance that offsets the potential risk, so cannot approve this aspect of the PR.  Your skills are a real asset to the WW3 community, so know that we appreciate your engagement.  Thank you in advance for understanding this decision was made with safety of the shared code base in mind.

MathieuDutSik:
> Hi @MathieuDutSik, I've kicked this around, and I'm not comfortable moving/removing existing initializations. Putting in initializations during development is not the same thing as removing them after the fact, especially with a large, community code base. I don't see any gain in this instance that offsets the potential risk, so cannot approve this aspect of the PR. Your skills are a real asset to the WW3 community, so know that we appreciate your engagement. Thank you in advance for understanding this decision was made with safety of the shared code base in mind.

It is a very simple choice: To assign or not a value to zero. If the field value is initialized inside of the code then the previous initialization does not make sense.

If you are so afraid of using arrays that have not been initialized to 0, why not put initialization in the code also for W3_NL2 / W3_NL3, etc.? The same issue would apply as well.

If my proposal were obviously wrong, it would show up immediately in the regression test. The DIA being wrong has an immediate impact on almost all tests.

As you know, there was an issue where I think the wrong choice was taken with respect to the ifdef (at least for the debugging tool, ifdef makes sense, for options like NL1, NL2 using input files actually gives you additional functionality like having different options for different grids). But at the very least I have to say that some reasons were given to me. I disagree but reasons were given and maybe possibly I can contribute to other aspects as there is more to WaveWatch III than ifdef.

Now, we are at another stage of action paralysis. Where exactly do you stand on the following:
1) "What about cleaning the code?" "Ah, but it could create problems we do not see"
2) "Removing unused variables" "Ah, but they could be used later"
3) "Eliminating memory alignment problems" "Ah, but this could break some systems that rely on those bugs"

It was a very simple PR and a very revealing one. Eventually, you will be afraid of everything and you will not be able to do anything.

MatthewMasarik-NOAA:
@MathieuDutSik I see no reason to remove these as there is no benefit from doing so.

MathieuDutSik:
> @MathieuDutSik I see no reason to remove these as there is no benefit from doing so.

Thank you for giving a reason for refusing the PR.

The problem of the `w3srcemd.F90` is that it is very complicated for unreasonable reasons. The variables for source term and diagonal terms should be set up in the relevant source terms code. Right now, I corrected only the W3_NL0 and W3_NL1 but the same correction can apply as well to other source terms.

The benefit of the PR is not in results. It is very marginally on performance (removal of a useless assignment to 0). But it is about code clarity. And please note that it decreases the number of ifdef, something that you used to see as a positive.

MatthewMasarik-NOAA:
@MathieuDutSik I agree that this PR has no affect on performance.  My opinion on the readability part though is that it makes the code less clear, not more.  My reasoning is that it provides a reference point for another programmer when trying to understand that subroutine, which they may not be familiar with.  With the assignments there, when building a mental picture of what happens during program execution, they will know those two variables are 0 up until their latter  assignment.  Without it they will not know what the value is, or if it is even defined.  This is my opinion, I respect that you may see this differently.  We have so far talked about readability.  Closely related is the "defensive programming" aspect.  A bug related to using those variables (possibly before assignment) will in most cases be easier to troubleshoot having been assigned a zero value, than a value determined by compiler implementation.  So taken together I see two negatives and no positives, and that's before we get to the fact it takes compute and man hours to test, which are finite resources.  This is why I am recommending against approval in it's current state.  Regarding removing ifdefs, I do think it was conveyed more than once that this is a careful, step-by-step process where we are removing the ifdefs that make sense to remove.  Removing variable initialization was not what I had in mind for the ifdef cleanup.  Not for this PR, and more importantly it is not a direction we want to go in for future PR's.  

That said I understand you may see some of these aspects differently.  Know that I appreciate the efforts you have been making to improve WW3, and look forward to working with you more toward that goal.

MatthewMasarik-NOAA (changes requested):
The typo correction in `model/src/w3snl1md.F90` is accepted.

The remove/move of variable initializations in `model/src/w3srcemd.F90` are rejected per the discussion thread above.

MatthewMasarik-NOAA:
Hi @MathieuDutSik.  There hasn't been any activity on this PR in a while, so we'd like to close it if there's no further work planned.  I'll leave it open for a few days in case you're interested, and will move to close it at the end of the week if not.  You're always welcome to reopen it anytime.  Additionally, if you'd like to discuss any aspects of this PR with me offline, I'd encourage you to do so,  matthew.masarik  noaa.gov.  I'd be happy to chat with you. Cheers

ukmo-ccbunney:
I'm going to stick my neck out and throw my two-penneth into the discussion here (just before it gets closed - sorry @MatthewMasarik-NOAA !)

Whilst I am not against initialising a variable twice for defensive purposes, I have noticed that this does have a significant impact in my work accelerating WW3 to run on GPUs. In this case, many variables (such as VSNL) need to have an extra seapoint dimension added to allow for parallelisation over multiple (thousands) of threads.

In this case, intialisation of these arrays suddenly has a non-negligable impact on execution time and is definitely not something that you want to be doing unnecessarily. There are lots of variables that are "zeroed" in W3SRCE (and in the indvidual source term routines), many probably unnecessarily. 

I'm not advocating that we need to do something about this now, but just thought I would mention it as it will come up again if my GPU work gains enough traction for me to put it in a PR! :)

MathieuDutSik:
Thank you @ukmo-ccbunney for reporting positive improvement for the removal of double inits.
If you feel it necessary, and of course, if there is an agreement, I can create a PR that addresses those double assignment problems.

ukmo-ccbunney:
> Thank you @ukmo-ccbunney for reporting positive improvement for the removal of double inits. If you feel it necessary, and of course, if there is an agreement, I can create a PR that addresses those double assignment problems.

Thanks @MathieuDutSik 
Following up from @MatthewMasarik-NOAA's sensible comments, perhaps it would be better if this sort of change is made in part of a larger piece of work (such as the GPU stuff I am working on), where it is much clearer there is a need to remove the double inits. I'm happy to tackle this at a later date!

MatthewMasarik-NOAA:
Thanks @ukmo-ccbunney  and @MathieuDutSik for your input! That's great information to know on GPU performance.  As @ukmo-ccbunney suggested, for now let's wait on any removal of double initializations as we work on a plan to systematically take care of these issues, while the GPU work progresses.  We're looking forward to continuing this conversation though, and we will figure out how to facilitate these improvements.

MatthewMasarik-NOAA:
Closing, due to no recent activity.  @MathieuDutSik you can of course re-open any time, as well as email so there are options should you choose.

Issue #684, opened 2022-04-22
Extend the mesh cap capabilities to unstructured meshes 


aliabdolali:
**Is your feature request related to a problem? Please describe.**
The unstructured meshes in WW3 are handled in two different ways: 1) the conventional carddeck, similar to the regular and curvilinear grids 2) the geographical domain decomposition using PDLIB switch. 

**Describe the solution you'd like**
The unstr mesh type could be an option for ww3_shel or ww3_multi, and the solver can be an option for the user.

**Additional context**
This might help the export and import grid to remain 1 but the type change to unstructured mesh where variable resolution is handled in the cap.

aliabdolali:
The following steps should be taken to achieve the milestone.

1. Extend the mesh cap to support unstructured meshes, Domain Decomposition and PLDIB library.
2. An unstructured mesh with identical nodes’ coordinates of 1 deg:
Corresponding to an existing atmwav test with the active FV3, but using cmeps and the current p8 physics
global rectilinear grid is generated (land included) to be tested with the meshcap.
3. Remove land from the mesh and check the meshcap.
4. Test mesh cap with variable resolution meshes (one of GEFSv13 or GFSv17 unstr options).
5. Test Mediator in the meshcap and efficiency check

PR #1002, opened 2023-04-26
Extend dev/ufs-weather-model NUOPC cap for use with WW3 unstructured meshes

DeniseWorthen:
# Pull Request Summary
 

Adds capability to run WW3 with unstructured meshes and domain decomposition in UFS

## Description

- Sets a flag (``unstr_mesh``) based on the WW3 variable ``gtype``
- Updates the Initialization of the mesh within the NUOPC cap to use the decomposition of the internal WW3 unstructured mesh if ``unstr_mesh`` is true.
- Utilizes the internal WW3 call to ``init_get_isea`` to fill import and export state variables over ``nseal_cpl`` points. The value of ``nseal_cpl`` (the number of PE-resident sea points) is identical to ``nseal`` for the structured case. For the unstructured case, ``nseal_cpl`` is the value of ``nseal`` excluding any ghost points.

### Issue(s) addressed

- fixes Issue https://github.com/ufs-community/ufs-weather-model/issues/1556

### Commit Message

Extend mesh-based NUOPC cap to unstructured WW3 meshes

### Check list  

### Testing

- New regression tests for UFS will be added using an unstructured mesh with domain decomposition
in UFS PR [1727](https://github.com/ufs-community/ufs-weather-model/pull/1727)


model/bin/switch_meshcap
```diff
@@ -1 +1 @@
-UWM NCO NOGRB DIST MPI OMPG OMPH PR3 UQ FLX0 SEED ST4 STAB0 NL1 BT1 DB1 MLIM FLD2 TR0 BS0 RWND WNX1 WNT1 CRX1 CRT1 O0 O1 O2 O3 O4 O5 O6 O7 O14 O15 IC0 IS0 REF0
```
JessicaMeixner-NOAA, line 1: Is there a reason the NCO switch was removed here?  We should also have that in the switch_meshcap_pdlib as well.

JessicaMeixner-NOAA (commented):
Thanks @DeniseWorthen for all of this great work!  I have one question on the switch file updates.

model/bin/switch_meshcap
DeniseWorthen: The NCO switch is used only in the ww3 standalone programs. It is not used when the mesh cap is run w/in UFS.

JessicaMeixner-NOAA: Thanks for the explanation.  We do need that switch for the pre and post processing jobs in the workflow and typically expect it to be in this switch file.  I will test to see if I can add it in the workflow without causing issues.

DeniseWorthen: I'm not sure how the ww3 codes would be compiled in UFS since they're not in the ``cmake/src_list.cmake``

JessicaMeixner-NOAA: They're not compiled with the ufs-weather-model, but the global-workflow uses the same switch file that the ufs-weather-model does to compile the pre-and post-executable. Actually going and looking at the code really quick, the exe for ww3_grid would need the NCO switch when creating the mod_def file, which is used for the forecast. So while again not strictly used for the forecast, I think it'd be simpler to leave this in.  That being said, I'll continue a test with the g-w taking this out of the forecast build and adding to pre- and post and see if it'll work or not.

JessicaMeixner-NOAA:
This PR will also fix https://github.com/NOAA-EMC/WW3/issues/684

model/bin/switch_meshcap
DeniseWorthen: This is ready to start on the UFS side, but it needs pre-approval first for the WW3 PR as a subcomponent. I can revert the removal of the NCO in the switch file if you require it.

JessicaMeixner-NOAA: @DeniseWorthen the testing on my end has gone a ltitle slow -- apologies.  My run from today ran out of memory.  If you do not want to wait for my testing, please revert to add back NCO. My preference is to revert even if this will work to add this in the global-workflow as it's just nice to have all the switches that we will need there and not have to add/subtract in other bulid scripts.  That being said, I'm continuing my tests and will keep you posted.

JessicaMeixner-NOAA (approved):
Thanks for adding NCO switches back in.  This simplifies things on the end to end system.  

Thanks you @DeniseWorthen  for this PR (and all the bug-fixes before this)!!! This is a huge effort and is greatly appreciated and is a huge step forward for us.  THANK YOU!!!

jkbk2004:
Tests are all completed at https://github.com/ufs-community/ufs-weather-model/pull/1727. Please, go ahead to merge the pr.

PR #1293, opened 2024-08-14
IC4M10: New Wave Attenuation Scheme

erinethomas:
# Pull Request Summary
 
A new wave attenuation scheme (IC4 method 10) based on Meylan Horvat Bitz Bennetts Ocean Modelling 2021

## Description
This PR adds a new wave attenuation scheme for IC4. It is based on Meylan Horvat Bitz and Bennetts, Ocean Modelling, 2021. It calculated wave damping based on scattering due to sea ice floes. It requires the use of the IS0 switch (no additional scattering term) and sea ice floe size diameter as input from ICECOEF5. 

Suggestions for a reviewers: 
Labels that should be added:   _new feature_

### Commit Message
IC4M10: New wave damping scheme in sea ice based on Meylan Horvat Bitz and Bennetts, Ocean Modelling, 2021. Co-Authors include: Erin Thomas, Cecilia Bitz, David Bailey, Nick Szapiro. 

### Check list  

- [X] Branch is up to date with the authoritative repository (NOAA-EMC) develop branch. 
- [X] Checked the [checklist for a developer submitting to develop](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-a-developer-submitting-to-develop). 
- [ ] If a version number update is required, checked the [updating version number](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-updating-version-number) checklist. 
- [ ] If a new feature was added, a regression test for testing the new feature is added. 
 
### Testing

* How were these changes tested?
These changes have been tested in fully coupled climate model simulations (E3SM and CESM) which have fully coupled wave (WW3) and sea ice components (the later of which contain sea ice floe size distribution).

* Are the changes covered by regression tests? (If not, why? Do new tests need to be added?)
* Have the matrix regression tests been run (if yes, please note HPC and compiler)?
* Please indicate the expected changes in the regression test output, (Note the [list](https://github.com/NOAA-EMC/WW3/wiki/How-to-use-matrix.comp-to-compare-regtests-with-develop#4-look-at-results) of known non-identical tests.)
* Please provide the summary output of matrix.comp (_matrix.Diff.txt_, _matrixCompFull.txt_ and _matrixCompSummary.txt_):

Issue #738, opened 2022-07-08
Improve numerical solution for waves in sea ice


cmbitz:
The numerical solution for wave attenuation in sea ice employs time splitting, which is only order dtg accurate. It is a noticeable error when dtg > 500 s or so. I've included a pdf to illustrate the error.

I recommend two solutions. (1) first don't use time splitting for the sea ice attenuation and instead include the sea ice "source" with the other sources and (2) employ subcycling on the global time step when the sea ice concentration is between say 5 and 95%. Happy to provide code.

Reducing dtg is a costly alternative.  

See attached pdf
[TimeSplittingProblem.pdf](https://github.com/NOAA-EMC/WW3/files/9073858/TimeSplittingProblem.pdf)

JessicaMeixner-NOAA:
@cmbitz if you have code, it would be great if you could make a PR with the update.  I can run any of the regression tests if needed.

aronroland:
I appreciate your contribution and I think that we really need to include the sea-ice source term in the sub-cycling and as well in the implicit scheme as source terms rather then within the splitting.

dabail10:
Hi all,

Here is a branch with Cecilia's proposed changes. I'm happy to discuss the best way to format these.

https://github.com/dabail10/WW3/tree/ic4method8

NickSzapiro-NOAA:
Thanks for sharing these developments, @dabail10 @cmbitz ! Do I understand correctly that the changes in model/src/w3srcemd.F90 are intended for all IC4 methods? Do you know if any other IC4M* developers have looked at these changes?

erinethomas:
It does appear to apply to all the IC4 methods. I have only tested it with one so far (the IC4M8 that @cmbitz helped develop) in E3SM.

NickSzapiro-NOAA:
Then maybe the next step is to make a PR, as @JessicaMeixner-NOAA suggested.

@dabail10 wrt the branch you shared, there are different implementations of the changes (if suitable) in
- model/src/w3fld1md.F90 with https://github.com/NOAA-EMC/WW3/pull/1188
- model/src/w3iorsmd.F90 with https://github.com/NOAA-EMC/WW3/pull/1224

Thanks @erinethomas! The IC4M8 you use will be "renumbered" as IC4M10 (in https://github.com/NOAA-EMC/WW3/pull/1211) since https://github.com/NOAA-EMC/WW3/pull/1176 is already merged

erinethomas:
[...]

PR #1176, opened 2024-01-17
R21B in IC4

ErickRogers:
# Pull Request Summary

Introduce IC4M8 and IC4M9 to WW3

## Description

This introduces IC4M8 and IC4M9 to WW3.
It should not change outcome of any WW3 runs which do not use the new features. (Reverse compatible.)
However, it does include two non-default options to aspects of IC4 that are not M8 or M9.
Namely:
* option for background dissipation at low frequencies
* option for using a larger number of steps in IC4M6 (was 10, and is now 16)

Some comments about "IC4 other than IC4M8/IC4M9 " are also changed/updated in the header of w3sic4md.F90.

Verbose descriptions of IC4M8 and IC4M9 can be found in header of w3sic4md.F90.

regtests to demo the new feature:
cd $REGTESTDIR
./bin/run_test -i input_IC4_M8 -w work_IC4_M8 $MODELDIR ww3_tic1.1
./bin/run_test -i input_IC4_M9 -w work_IC4_M9 $MODELDIR ww3_tic1.1

ER: I need to describe the feature in the manual. I added it to my to-do list.

* Suggested reviewers: @JessicaMeixner-NOAA , @MatthewMasarik-NOAA 
* Labels: new feature
* Changes are not expected from this PR.

### Issue(s) addressed
This is for issue #1167

### Commit Message
Introduce IC4M8 and IC4M9 to WW3

### Check list  

- [x] Branch is up to date with the authoritative repository (NOAA-EMC) develop branch. 
- [x] Checked the [checklist for a developer submitting to develop](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-a-developer-submitting-to-develop). 
- [x] If a version number update is required, checked the [updating version number](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-updating-version-number) checklist. 
- [x] If a new feature was added, a regression test for testing the new feature is added. 
 
### Testing

* How were these changes tested? 2 new regression tests were added, see here: https://github.com/NOAA-EMC/WW3/commit/d4640abf74bb4e8a63b4a9a69d502044ba885812
* Are the changes covered by regression tests? Yes, see prior bullet.
* Have the matrix regression tests been run (if yes, please note HPC and compiler)? No.
* Please indicate the expected changes in the regression test output. No changes expected.
* Please provide the summary output of matrix.comp (_matrix.Diff.txt_, _matrixCompFull.txt_ and _matrixCompSummary.txt_): N/A. Matrix was not run.

PR #1294, opened 2024-08-15
Updated Numerical Solution for Wave Attenuation in sea ice

erinethomas:
# Pull Request Summary
This PR provides code modifications to apply a new MISC namelist parameter called "ICNUMERICS". Setting ICNUMERICS to be "TRUE"  allows the user to place the 'sea ice' source terms with the other source terms, thus reducing the errors introduced with the time splitting method for wave attenuation in sea ice resulting in more accurate wave action in sea ice.

## Description
The default behavior for wave attenuation in sea ice uses 'time splitting' for the wave action equation. This PR and the use of the new namelist parameter places the 'sea ice' source terms with the other source terms, thus reducing the errors introduced with the time splitting method for wave attenuation in sea ice resulting in more accurate wave action in sea ice. 

Currently, this code is intended to be used in tandem with the newly proposed Meylan et al 2021 attenuation scheme : IC4M10 (see PR: https://github.com/NOAA-EMC/WW3/pull/1293 ). This combination (IC4M10 plus ICNUMERICS=TRUE) has been tested in coupled CESM and E3SM runs, yet we note, at this time, this new numerics calculation for wave attenuation has **not** yet been tested with any other IC switch settings or other IC4 methods (1-9), although, we expect similar improvements for all IC methods. 

This PR is the original work of @cmbitz and Vince Cooper (University of Washington). Collaborators include @erinethomas, @dabail10, @NickSzapiro-NOAA

### Issue(s) addressed
This PR fixes the following Issue:  https://github.com/NOAA-EMC/WW3/issues/738 

### Commit Message
Introduction of a new MISC namelist parameter (ICNUMERICS) that allows for more accurate numerics scheme for wave attenuation in sea ice. Coauthors: Cecilia Bitz, Vince Cooper, Erin Thomas, David Bailey, Nick Szapiro.

### Check list  

- [X] Branch is up to date with the authoritative repository (NOAA-EMC) develop branch. 
- [X] Checked the [checklist for a developer submitting to develop](https://github.com/NOAA-EMC/WW3/wiki/Code-Management#checklist-for-a-developer-submitting-to-develop). 
 
### Testing

This code has been tested by @cmbitz (details to be posted below), by @erinethomas in E3SM configurations with wave-sea ice coupling, and by @dabail10 in CESM configurations with wave-sea ice coupling. 


NickSzapiro-NOAA:
Maybe it is possible to progress with some discussion/review during this pause, @MatthewMasarik-NOAA ? I don't know if any IC4 developers in particular have feedback

MatthewMasarik-NOAA:
Hi @NickSzapiro-NOAA, others should feel free to discuss.  If this is not needed for either gfsv17 or gefsv13 though, than Avichal has reiterated that the waves team needs to pause any reviewing since we don't have the manpower currently.

MatthewMasarik-NOAA:
@erinethomas, we will also process this in ~November.  Thank you for the very informative PR header.

cmbitz:
I've made a github repository about this issue at https://github.com/cmbitz/WW3Numerics where I've written up some more detailed notes about this issue and show some results in slides from a number of integrations to test it. I also provide a Python Notebook with some idealized examples that can be solved exactly where I compare the various solution options. In short, the modifications I'm suggesting can be implemented so they only matter in the sea ice. They can work for any IC/IS option combination, though I confess I've only tested with IC4Method10. I'm 100% sure it would improve wave amplitudes in the sea ice for any IC option. Happy to talk to anyone who wants to listen. I understand that this issue is on the back-burner, which is okay with me. I'll wait.

erinethomas:
thanks for the extra details @cmbitz!

sbanihash:
Thank you @erinethomas for your work on this pull request and your patience as we were handling other tasks during these past few months. I have reviewed the PR and tested the changes to make sure it does not change outcome of any WW3 runs not using this new feature. Everything looks good but since this update will add a numerical solution to the wave attenuation, I would like to ask if you could maybe follow the steps in a similar PR (https://github.com/NOAA-EMC/WW3/pull/1176) and document this new change in the WW3 manual section. If that's not possible at this moment we appreciate it if you could open an [issue](https://github.com/NOAA-EMC/WW3/issues) and assign it to yourself so that you could add it whenever you are ready to do so. I already see a nice link to the document that @cmbitz has added to this PR . Another suggestion is that since these new changes can not be tested in the current ww3 regression tests, it would be great if you could take a look at PR 1176 and see how a regression test is added to that PR and try making a similar test case for your changes. The test could be an IC4-M10 case inside the ww3_tic1.1 and besides checking the steps that can be tracked in [previously mentioned PR](https://github.com/NOAA-EMC/WW3/pull/1176/files#diff-3589e2082baa9eea5303fe7fce9872cdd1476064878b1363b6d9b54cf0afea08), you could also reach out to me as well and I would be more than happy to help set up this test case with you. Thanks!

model/src/w3srcemd.F90
```diff
@@ -2097,7 +2145,7 @@ SUBROUTINE W3SRCE ( srce_call, IT, ISEA, JSEA, IX, IY, IMOD,          &
 #ifdef W3_IC3
         ATT=EXP(ICE*VDIC(IS)*DTG)
 #endif
-#ifdef W3_IC4
+#ifndef W3_IC4_ACCURATE_NUMERICS
```
sbanihash: In a case of IC0 for example, where there is no IC1-IC5 defined, then VDIC has not been defined (look at line 744 in this same script) and the build will give this error: This name does not have a type, and must have an explicit type.   [VDIC]
        ATT=EXP(ICE*VDIC(IS)*DTG)
suggested changes is:
#ifdef WW3_IC4
#ifndef WW3_IC4_ACCURATE_NUMERICS
        ATT = EXP(ICE * VDIC(IS) * DTG)
#endif
#endif

sbanihash (changes requested):

model/src/w3srcemd.F90
sbanihash: @erinethomas @NickSzapiro-NOAA do you agree with this change? If so could you please go ahead and make the changes?

NickSzapiro-NOAA: Keeping both makes sense to me to protect VDIC

ErickRogers:
Why is this implemented as a switch? Could a logical variable in the IC4 namelist do the same thing?

dabail10:
> Why is this implemented as a switch? Could a logical variable in the IC4 namelist do the same thing?

Hi Erick. We were following the example of the W3_IC? options here. Happy to change it to a namelist option if this is preferred.

ErickRogers:
Hi David.
Yes, I'm not sure if this is covered in the guidelines for developers, but the switches tend to be used for things that are pretty major, like selecting a propagation scheme, selecting a physics parameterization, selecting parallel computing method, or compiling in extra software (e.g., SCRIP). When a switch is changed, it requires a re-compile, whereas changing a namelist variable in the ww3_grid.inp file does not require a re-compile. Any new switch needs to be accommodated in the build system (/bin/build_utils.sh), which is an argument for "fewer switches=better". If this is limited to IC4, you can use a new logical variable in the IC4 namelist. If your option adds to the computation time, e.g., by forcing a small time step, then the original setting should be the default. (This will also make the discussion re: NCEP's "matrix" testing straightforward.) If this is intended to be used with any of the other IC# methods, you can put the new logical variable in the MISC namelist instead of the IC4 namelist.

dabail10:
Ok. I will work with @erinethomas on this.

ErickRogers:
Adding a namelist variable is very easy. I just did this the other day, to put in a new option for hacking the ice thickness. You can look at an existing variable to see how it is done, e.g. "grep IC4 *.F90 | grep -i ki". Any new variable will have 2 names. One is used in ww3_grid.inp and the other is used internally, e.g. IC4KI and IC4_KI. There are 3-4 files that need minor editing to add the new variable, as you will see with the output from above grep command. Keep in mind that after making this change, any old mod-def files that you have will no longer work.

erinethomas:
This seems like a good way forward (to go with namelist options rather than switches) - I will update the PR as soon as I can!

sbanihash:
@erinethomas @dabail10 Do we have any updates to this PR?

erinethomas:
yes -I  have worked out how to move this from a switch setting to a namelist parameter - I am still doing some final tests - code coming soon.

erinethomas:
@cmbitz @dabail10 @NickSzapiro-NOAA 
UPDATE: I have modified the code to use a logical Namelist setting called "ICNUMERICS"/"IC_NUMERICS". This namelist setting should be placed under "MISC" namelist category. I placed it under MISC (as opposed to IC4) since, in the future this numerical scheme should be applicable (and tested) for any Sea Ice dissipation scheme (not just IC4). This code  **should** be applicable for users to use in any IC/IS combination. Keeping in mind this flexibility, would you all take a look at the modifications, and check if the logic I implemented is correct ... especially with respect to the changes I made in w3srcemd.F90 and any "IC/IS" switches used.... (for example Line ~1330, Line ~1416, Line 1460, etc.)

sbanihash:
Thank you @erinethomas for your updates! Please let me know when you are ready for the second round of review.

ErickRogers (approved):
The recent updates by Erin look good to me. I don't have any other concerns or suggestions. I suppose it needs to get through the matrix testing now.

sbanihash:
Thank you @ErickRogers for your review. @erinethomas I will be starting my second round of review soon. Please let me know if there are still changes or checks you need to do on you side before I do so. Thanks

erinethomas:
thanks @ErickRogers, @sbanihash - you are welcome to review this now - but I would very much like one or more of the other developers (@cmbitz, @dabail10, @NickSzapiro-NOAA) to review these changes as well before the PR is finalized.

model/src/w3gridmd.F90
```diff
@@ -809,6 +809,7 @@ MODULE W3GRIDMD
   !
   REAL(8)                 :: GSHIFT ! see notes in WMGHGH
   LOGICAL                 :: FLC, ICEDISP, TRCKCMPR
+  LOGICAL                 :: ICNUMERICS
```
dabail10, line 812: This is a nit-picky thing, but do we need both ICNUMERICS and IC_NUMERICS? I feel like the other ice flags have IICESMOOTH and ICESMOOTH.

model/src/w3srcemd.F90
```diff
@@ -1423,6 +1457,11 @@ SUBROUTINE W3SRCE ( srce_call, IT, ISEA, JSEA, IX, IY, IMOD,          &
 #ifdef W3_UOST
         VS(IS) = VS(IS) + VSUO(IS)
 #endif
+        IF ( IC_NUMERICS .AND. ICE.GT.0. ) THEN
+#if defined(W3_IC1) || defined(W3_IC2) || defined(W3_IC3) || defined(W3_IC4) || defined(W3_IC5)
+          VS(IS) = VS(IS) + VSIC(IS)
```
dabail10, line 1462: In my version of the code we also have VD(IS) = VD(IS) + VDIC(IS)?

dabail10: Oh wait, I see this is down further.

model/src/w3gdatmd.F90
```diff
@@ -700,6 +701,7 @@ MODULE W3GDATMD
 
     LOGICAL          :: GINIT, FLDRY, FLCX, FLCY, FLCTH, FLCK, FLSOU, IICEDISP,&
          IICESMOOTH
+    LOGICAL, POINTER :: IC_NUMERICS
```
dabail10: Is this supposed to be a pointer in both places?

dabail10 (commented):
Testing this out in CESM.

model/src/w3gridmd.F90
erinethomas: I was following the naming convention for "IC4KI and IC4_KI" - but the MISC namelist settings seem to have the 'extra I' rather than the underscore... we could probably pick either convention. I have no strong opinion on this.

model/src/w3gdatmd.F90
erinethomas: I'm not 100% sure to be honest - I don't understand why some of the namlist settings here are pointers... but other are not... for example, most of the IC4 namelist settings are pointers in both places? but the IICEDISP is only a pointer in one place?

model/src/w3srcemd.F90
```diff
@@ -1454,6 +1498,11 @@ SUBROUTINE W3SRCE ( srce_call, IT, ISEA, JSEA, IX, IY, IMOD,          &
 
       !
       DT     = MAX ( 0.5, DT ) ! The hardcoded min. dt is a problem for certain cases e.g. laborotary scale problems.
+      IF ( IC_NUMERICS .AND. ICE .GT. 0.01 .and. ICE .LT. 0.95) THEN
```
NickSzapiro-NOAA: ICE.GT.0 is used elsewhere. Is this a big change in efficiency or to avoid something in particular? 

If not, I'm hesitant as these seem buried and may lead to spatial differences around these sea ice concentrations

erinethomas: Good question @NickSzapiro-NOAA ... @cmbitz can you explain your logic for putting 'ICE.GT.01' rather than 0 here?

dabail10:
@cmbitz was actually advocating that we should remove this timestep limitation under the ice. However, I don't know if that is only when OFFSET=0.5 in w3srcemod.F90

model/src/w3gridmd.F90
ErickRogers: It is fine, IMO.

dabail10:
These changes work in the CESM code.

model/src/w3srcemd.F90
cmbitz: It doesn't really matter if it is 0 or 0.01 provided tiny dots of sea ice don't hang around. I was just trying to make the code run a bit faster when the sea ice concentration is small since there isn't much point in increasing the accuracy for less than 1% of the area. The error in the 99% or more area of open water will dominate anyway. Though @dabail10 is right that I think we should not bother with this line of code at all. I wouldn't bother with it regardless  of OFFSET. We had limited the timestep since there are conditions where my recommended numerical changes in the sea ice are less accurate unless the timestep is also reduced. However, I don't think it is that important. What is much more important is the "merging" of the sea ice source with the other source terms that is the main thrust of this pull request.

erinethomas: thanks for the extra details! I will remove this line and update the PR.

sbanihash:
@erinethomas are you done with the changes that you were planning on making? Is the PR ready for testing again?

erinethomas:
@sbanihash - I believe this PR is ready for review again.

model/src/w3srcemd.F90
erinethomas: these lines have been removed.

sbanihash:
> yes -I have worked out how to move this from a switch setting to a namelist parameter - I am still doing some final tests - code coming soon.

@erinethomas could you edit your PR description to reflect this change?

model/src/w3gdatmd.F90
sbanihash: @erinethomas @dabail10 While trying to test this PR, my runs failed in ww3_grid. After debugging I figured that the IC_NUMERICS should be just a logical in line 704 and pointer reference on line 1221. I changed the code accordingly and it got passed that error. I have a hard time understanding why the pointers should be defined like this but I can confirm that without this change the code does not run on at least the machine I was trying to run it on (Hera).
@erinethomas you did mention the IC4 namelists, but I guess since they are only complied within an #ifdef, they don't have to be defined the same way as MISC namelists? (or at least that's the only difference I see between IC_NUMERICS or IICEDISP and IC4 namelists.)

erinethomas: @sbanihash - ah I see. perhaps that is why. I will update the PR to reflect these changes (logical on line 704, pointer on line 1221)

sbanihash: Thank you @erinethomas. Do you get the same error when running? (I mean without this change?) because I see in the discussions that @dabail10 says changes work in CESM code.

sbanihash:
@erinethomas just double checking to see if you have that one last change ready to be committed for me to submit my final review? Thank you in advance!

erinethomas:
@sbanihash - thanks for the reminder - change to line 704 (removed the pointer) has now been made. Review ready.

sbanihash (approved):
I have reviewed the changes in this PR and tested the regression tests successfully. This PR will write an additional term in all ww3_grid.out files for ICNUMERICS and result in ww3_grid.out files being different for all regression tests. For that reason the regtests were first ran without any changes in the original PR and the tests were successful. In an attempt to compare matrixes, the write statements in ww3gridmd.F90 were taken out before the second round of regression tests and the matrix diff result is as below:
mww3_test_03/./work_PR2_UNO_MPI_e                     (1 files differ)
mww3_test_03/./work_PR2_UQ_MPI_d2                     (16 files differ)
mww3_test_03/./work_PR3_UQ_MPI_e_c                     (1 files differ)
mww3_test_03/./work_PR2_UNO_MPI_d2                     (16 files differ)
mww3_test_03/./work_PR3_UQ_MPI_d2_c                     (15 files differ)
mww3_test_03/./work_PR2_UQ_MPI_e                     (1 files differ)
mww3_test_03/./work_PR3_UNO_MPI_e_c                     (1 files differ)
mww3_test_03/./work_PR1_MPI_d2                     (11 files differ)
mww3_test_03/./work_PR3_UNO_MPI_d2_c                     (15 files differ)
mww3_test_03/./work_PR3_UNO_MPI_d2                     (16 files differ)
mww3_test_03/./work_PR3_UQ_MPI_d2                     (15 files differ)
mww3_test_09/./work_MPI_ASCII                     (0 files differ)
ww3_tp2.6/./work_ST4_ASCII                     (0 files differ)
ww3_ufs1.3/./work_a                     (2 files differ)

All diff's are as expected. I approve this PR. 
[matrixCompSummary.txt](https://github.com/user-attachments/files/19474755/matrixCompSummary.txt)
[matrixCompFull.txt](https://github.com/user-attachments/files/19474756/matrixCompFull.txt)

[matrixDiff.txt](https://github.com/user-attachments/files/19474758/matrixDiff.txt)

sbanihash:
@erinethomas I had already made that change and ran the tests, so when I double checked your latest commit and made sure it is the same as the change I had made in my runs, I went ahead and approved this PR. Thank you again and also thanks to @dabail10, @cmbitz and @NickSzapiro-NOAA for your work on this PR and to @ErickRogers for his review. 
@erinethomas I will be creating a pull request next week with the ICNUMERICS truned on in IC4M10 so that this new feature is tested in our regtests as well. Stay tuned!

sbanihash:
@erinethomas @cmbitz I have been working on adding a regtest for this feature, by changing ICNUMERICS=T in the input for the IC4M10 case in tic1.1 [,](https://github.com/NOAA-EMC/WW3/tree/develop/regtests/ww3_tic1.1/input_IC4_M10) but when comparing the results to a case where ICNUMERICS=F, I do not see any changes in output. I know you had mentioned that your tests have been successful, could you confirm that what is already in the IC4_M10 case is sufficient to test this numerical solution option?
Thanks

dabail10:
These changes should only impact ice covered areas. Do you have coupling with the sea ice model?

sbanihash:
Thanks @dabail10 for your prompt response. So you are confirming that we would need to test it in a coupled setting to see the results? If so I would have to look at setting up a test within the ufs-weather-model setup.

dabail10:
You should definitely test it in the coupled setup. Or do you have a method in WW3 to read in sea ice data?

sbanihash:
@dabail10 yes, we read an ice input for example this is how the ice is being read in the regtest that I am working on:
https://github.com/NOAA-EMC/WW3/blob/0a9e10f7d55f2b73eb2ffc03a1dd05fd05f434c7/regtests/ww3_tic1.1/input_IC4_M10/ww3_prep_icecon.inp#L34

NickSzapiro-NOAA:
@sbanihash For coupled test, I have stale PR (https://github.com/ufs-community/ufs-weather-model/pull/2072) once this is synced to dev/ufs-weather-model. 

Idea is to switch on wave-ice coupling in cpld_control_pdlib_p8 and cpld_debug_pdlib_p8 RTs

sbanihash:
Thanks @NickSzapiro-NOAA was just looking at your PR. Let me see if I can do a test within your branch with the addition of the changes in this PR. I was under the impression that as long as I am reading an ice input, turing on ICNUMERICS=T should result in answer changes even in a  WW3 regtest, am I wrong?

dabail10:
I agree. If you are reading in a sea ice field, then the numerics should be doing something in the presence of ice.

NickSzapiro-NOAA:
Do you have plots for your ww3 regtest? Are there waves in the ice?

sbanihash:
@NickSzapiro-NOAA I can definitely generate some and share.

sbanihash:
@NickSzapiro-NOAA I have opened an issue of what I think to be the problem with this PR not working as expected when turning IC_NUMERICS on + a PR with the fixes which worked for the test case I have added. Still need to wait for all the regtests to finish to have the full matrix comparisons so that I could officially have the PR reviewed and tested by others.

Issue #1386, opened 2025-03-07
IS1 not contributing to action density spectrum


Biao-Zhao:
In w3srcemd.F90, it appears that the ice scattering source term calculated by **IS1** is not contributing to the increment of the action density spectrum (**SPEC**).

![Image](https://github.com/user-attachments/assets/0e6b1c06-db82-402d-b74b-d5e76d4cab91)

![Image](https://github.com/user-attachments/assets/630d1f04-9b98-434d-aedc-dc8afaacd645)

From the two screenshots above, we can see that **ATT** expects VDIR from the call to **W3SIS1**. However, **W3SIS1** does not return it.

From an energy conservation perspective, even if **W3SIS1** were to return **VDIR**, I believe it would still be inappropriate to consider the ice scattering term as dissipation. Based on my understanding, it might be better to redistribute **SPEC** in different directions, following the same approach as **IS2**. Alternatively, we could include **VSIR** in **VS**, such as:

**VS(IS) = VS(IS) + VSIR(IS)**

Anyway, I might have misunderstood some details, and I appreciate any corrections or clarifications. Please feel free to share your thoughts.

NickSzapiro-NOAA:
@cmbitz reports the same in https://github.com/NOAA-EMC/WW3/issues/1305

Biao-Zhao:
> [@cmbitz](https://github.com/cmbitz) reports the same in [#1305](https://github.com/NOAA-EMC/WW3/issues/1305)

Hi Nick,

Thanks for sharing the information. I hadn't noticed it before. Personally, I’m not particularly interested in wave-ice interaction, but I came across this bug while reviewing

>  w3srcemd.F90

As @cmbitz  said, it is easy to fix it. I shared this here as a reminder for anyone working on this topic.  Since scattering doesn’t dissipate wave energy, I believe the correct solution is to redistribute the scattered energy to other directions, similar to IS2.  In this sense, adding the following line should solve the problem

`SPEC(1+(IK-1)*NTH:NTH+(IK-1)*NTH) = SPEC(1+(IK-1)*NTH:NTH+(IK-1)*NTH) +  VSIR(IS) * DTMIN`

Of course,  the following line must be commented out accordingly. Not only because it doesn’t work, but also for physical considerations

    `!/IS1             ATT=ATT*EXP(ICE*VDIR(IS)*DTG)`

If anyone thinks it’s necessary, I would be happy to submit a pull request

sbanihash:
[...]

Issue #1305, opened 2024-10-02
Wave scattering in sea ice incorrectly coded for IS1 option


cmbitz:
**Describe the bug**
The  IS1 option calls a subroutine that returns a source term VSIR, but VSIR is never used. It does not return a "diagonal" term VDIR, though VDIR would be correctly used if it had. The relevant code is
```
    CALL W3SIS1 ( SPEC, ICE, VSIR )
    ATT=ATT*EXP(ICE*VDIR(IS)*DTG)
```

**To Reproduce**
Do a run where W3_IS1 is defined and compare it to a run without defining W3_IS1. The output will be bit-for-bit the same.

**Expected behavior**
What should happen is that IS1 scattering dampens wave energy in the sea ice.

**Screenshots**
N/A

**Additional context**
I haven't tried to fix this, but I know how to do it. I'm not interested in using this option myself. However, I think some people do use IS1, and may not realize it  does nothing. I'd like to also mention this bug is unrelated to the other issue (#1294) that we have about numerical solutions involving sea ice. #1294 is not a bug, but rather a recommendation to improve the numerics. This is plainly a bug. I'd be happy to provide a pull request with this fix combined with correction to the numerics that I am recommending in issue #1294

PR #1387, opened 2025-03-07
Update w3srcemd.F90 for considering the contribution of ice scattering of IS1

Biao-Zhao:
I reported a bug [#1386](https://github.com/NOAA-EMC/WW3/issues/1386), the current IS1 is not contributing to action density spectrum. Since scattering doesn’t dissipate wave energy, I believe the correct solution is to redistribute the scattered energy to other directions, similar to IS2.

# Pull Request Summary
 

### Issue(s) addressed
Addresses #1305
Addresses #1386

### Check list  

### Testing

### DIFF
--- a/model/src/w3srcemd.F90
+++ b/model/src/w3srcemd.F90
@@ -2127,7 +2127,7 @@ CONTAINS
 
 #ifdef W3_IS1
         ! IS1 doesn't dissipate wave energy, it just redistribute the scattered energy to other directions
-        SPEC(1+(IK-1)*NTH:NTH+(IK-1)*NTH) = SPEC(1+(IK-1)*NTH:NTH+(IK-1)*NTH) + VSIR(IS)*DTMIN
+        SPEC(1+(IK-1)*NTH:NTH+(IK-1)*NTH) = SPEC(1+(IK-1)*NTH:NTH+(IK-1)*NTH) + VSIR(IS)*DTG
 #endif
 
 #ifdef W3_IS2
--- a/model/src/w3srcemd.F90
+++ b/model/src/w3srcemd.F90
@@ -2124,6 +2124,12 @@ CONTAINS
         !
         ! Second part of ice term integration: scattering including re-distribution in directions
         !
+
+#ifdef W3_IS1
+        ! IS1 doesn't dissipate wave energy, it just redistribute the scattered energy to other directions
+        SPEC(1+(IK-1)*NTH:NTH+(IK-1)*NTH) = SPEC(1+(IK-1)*NTH:NTH+(IK-1)*NTH) + VSIR(IS)*DTMIN
+#endif
+
 #ifdef W3_IS2
         IF (IS2PARS(2).GE.0) THEN
           IF (IS2PARS(20).GT.0.5) THEN


sbanihash:
@Biao-Zhao I have started reviewing this PR and looking at the discussion that has happened in #1386, I noticed you also mentioned that this part of the code should be commented out:

ATT=ATT\*EXP(ICE*VDIR(IS)*DTG)

Is there a reason you did not take that into account in your PR?

Biao-Zhao:
> @Biao-Zhao I have started reviewing this PR and looking at the discussion that has happened in #1386, I noticed you also mentioned that this part of the code should be commented out:
> 
> ATT=ATT*EXP(ICE*VDIR(IS)*DTG)
> 
> Is there a reason you did not take that into account in your PR?

Hi @sbanihash, yes, I didn't comment it out because W3sis1md.F90 doesn’t return VDIR.
The subroutine is defined as 

`SUBROUTINE W3SIS1 (A, ICE, S), `

so VDIR stays 0 as initialized in W3srcemd.F90. In this case, whether that line is commented out or not doesn't make a difference. Personally, I think this line “ ATT=ATT*EXP(ICE*VDIR(IS)*DTG)”  should be commented out. Since when IS1 is switched on, the wave energy is just redistributed rather than dissipated.

Hope that makes sense.

sbanihash:
@Biao-Zhao I have taken a closer look at the original issues that have been listed in this PR, your specific changes to address them and the differences between IS1 and IS2 that are documented in the WW3 manual.
Although I agree that the issues which are mentioned in this PR are valid issues(#1305 & #1386 ) I do not fully understand how your PR is a fix to those specifically. 
I believe that your PR will treat IS1 in a completely new way that has not been the origin of the method in WW3. It could still be a great way to account for the effects of ice scattering that could be named a new method and documented.
I would feel more confident if to address the issues, we just fix these lines, since apparently it's not correct : 

https://github.com/NOAA-EMC/WW3/blob/0a9e10f7d55f2b73eb2ffc03a1dd05fd05f434c7/model/src/w3srcemd.F90#L2178

And then if you would like to proceed with your method, you could open a new PR, fully test your suggested method and document it for future reference.

For more details, this is my take on IS1 and IS2 and I would be more than happy if someone with more expertise could chime in and correct me if I am misunderstanding anything:
IS1 and IS2 are both considered scattering methods, IS1 is more basic. Referring to the WW3 manual: At each discrete frequency and direction IS1 will reduce the wave energy by the amount of \alpha and redistribute it to all
direction in the same discrete frequency to conserve energy (which is done in w3sis1md.F90), while IS2 is a Floe-size dependent scattering and dissipation which is a more physics-based scattering.

@ErickRogers I would appreciate your review on this PR as well. I might be missing something here and any input is appreciated.

jjimenez233:
@sbanihash,

I inspect PR and no problem with code for   "Vdir".   Not return from subroutine.  Author tell you this, he give right answer.  Trust author or maybe just back to school, then you also be expert some day!

<div font="Helvetica" align="center">
&NBSP&NBSP  "Jey" Jimenez
</div>

ErickRogers:
I agree with the prior comments that it looks broken. The routine computes VSIR but then it wants to apply VDIR rather than VSIR. Presumably this was working some time in the past. My guess is that it became broken when the "ATT=ATT*\EXP(ICE...)" stuff was added to all the ice routines. And I agree that this form "ATT=ATT*EXP(ICE...." is for dissipation rather than redistribution. FYI Stefan is : <stefan.zieger@bom.gov.au> (or at least was in 2022), but I doubt that he is interested in this now. @Biao-Zhao if I understood your message, you are not planning to apply this routine, but rather you are just reporting a bug that you found while scanning the code. If we can't identify anyone interested in using IS1, I wonder if we should just deprecate it....remove it entirely, or mark it as broken, with error message, etc. I don't think we should try to fix it unless someone is volunteering to document/test/illustrate the fix.

Biao-Zhao:
Yes, I don't focus on the sea ice and related IS1.  I came across  this bug  while going though the source term driver codes. I support the treatment proposed by @ErickRogers .  If we don't fix this bug, it could mislead users who use IS1 into thinking that ice scattering doesn't work

PR #1396, opened 2025-03-17
Fix a typo which breaks the build of the manual.

pvc1989:


sbanihash (approved):
Changes look good. Thank you @pvc1989 for catching the typo and submitting a PR for the fix.

PR #1402, opened 2025-04-04
Fix zero boundary and orbital velocity issue in the cap

danishyo:
Original issue can be found at https://github.com/oceanmodeling/ufs-weather-model/issues/145#issuecomment-2640891791
It cause zero values along the boundary for exchange variables (like SXX) from cap.
Also, original code will derive weird orbital velocity (looks like domain-decomposition map, like the following figure)
image

@uturuncoglu , and @janahaddad are aware of these issues. 
@DeniseWorthen , could you have a look on this fix? Thx!

### DIFF
--- a/model/src/wav_import_export.F90
+++ b/model/src/wav_import_export.F90
@@ -680,7 +680,7 @@ contains
         call init_get_isea(isea, jsea)
         ix  = mapsf(isea,1)
         iy  = mapsf(isea,2)
-        if (mapsta(iy,ix) /= 0 .and. HS(jsea) > zero .and. &
+        if (mapsta(iy,ix) == 1 .and. HS(jsea) > zero .and. &
             sqrt(USSX(jsea)**2+USSY(jsea)**2)>zero .and. sqrt(USSHX(jsea)**2+USSHY(jsea)**2)>zero ) then
            sww = atan2(USSHY(jsea),USSHX(jsea)) - UD(isea)
            alphal = atan( sin(sww) / (                                       &
@@ -708,7 +708,7 @@ contains
          isea = iaproc + (jsea-1)*naproc
          ix  = mapsf(isea,1)
          iy  = mapsf(isea,2)
-         if (mapsta(iy,ix) /= 0) then
+         if (mapsta(iy,ix) == 1) then
             ! note: an arbitrary minimum value of 0.2 is set to avoid zero
             !       Langmuir number which may result from zero surface friction
             !       velocity but may cause unphysically strong Langmuir mixing
@@ -729,7 +729,7 @@ contains
         call init_get_isea(isea, jsea)
         ix  = mapsf(isea,1)
         iy  = mapsf(isea,2)
-        if (mapsta(iy,ix) /= 0) then
+        if (mapsta(iy,ix) == 1) then
           sw_ustokes(jsea) = USSX(jsea)
         else
           sw_ustokes(jsea) = 0.
@@ -744,7 +744,7 @@ contains
         call init_get_isea(isea, jsea)
         ix  = mapsf(isea,1)
         iy  = mapsf(isea,2)
-        if (mapsta(iy,ix) /= 0) then
+        if (mapsta(iy,ix) == 1) then
           sw_vstokes(jsea) = USSY(jsea)
         else
           sw_vstokes(jsea) = 0.
--- a/model/src/wav_import_export.F90
+++ b/model/src/wav_import_export.F90
@@ -680,7 +680,7 @@ contains
         call init_get_isea(isea, jsea)
         ix  = mapsf(isea,1)
         iy  = mapsf(isea,2)
-        if (mapsta(iy,ix) == 1 .and. HS(jsea) > zero .and. &
+        if (mapsta(iy,ix) /= 0 .and. HS(jsea) > zero .and. &
             sqrt(USSX(jsea)**2+USSY(jsea)**2)>zero .and. sqrt(USSHX(jsea)**2+USSHY(jsea)**2)>zero ) then
            sww = atan2(USSHY(jsea),USSHX(jsea)) - UD(isea)
            alphal = atan( sin(sww) / (                                       &
@@ -708,7 +708,7 @@ contains
          isea = iaproc + (jsea-1)*naproc
          ix  = mapsf(isea,1)
          iy  = mapsf(isea,2)
-         if (mapsta(iy,ix) == 1) then
+         if (mapsta(iy,ix) /= 0) then
             ! note: an arbitrary minimum value of 0.2 is set to avoid zero
             !       Langmuir number which may result from zero surface friction
             !       velocity but may cause unphysically strong Langmuir mixing
@@ -729,7 +729,7 @@ contains
         call init_get_isea(isea, jsea)
         ix  = mapsf(isea,1)
         iy  = mapsf(isea,2)
-        if (mapsta(iy,ix) == 1) then
+        if (mapsta(iy,ix) /= 0) then
           sw_ustokes(jsea) = USSX(jsea)
         else
           sw_ustokes(jsea) = 0.
@@ -744,7 +744,7 @@ contains
         call init_get_isea(isea, jsea)
         ix  = mapsf(isea,1)
         iy  = mapsf(isea,2)
-        if (mapsta(iy,ix) == 1) then
+        if (mapsta(iy,ix) /= 0) then
           sw_vstokes(jsea) = USSY(jsea)
         else
           sw_vstokes(jsea) = 0.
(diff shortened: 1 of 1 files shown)


model/src/wav_import_export.F90
```diff
@@ -744,7 +744,7 @@ subroutine export_fields (gcomp, rc)
         call init_get_isea(isea, jsea)
         ix  = mapsf(isea,1)
         iy  = mapsf(isea,2)
-        if (mapsta(iy,ix) == 1) then
+        if (mapsta(iy,ix) /= 0) then
```
DeniseWorthen: Please don't change anything w/in the CESMCOUPLED ifdefs unless you're testing in CESM and know that the change is acceptable.

danishyo: Got it! Thanks for the notice!

sbanihash: @danishyo can you make sure that this requested change is reflected in this PR?

danishyo: @sbanihash I didn't intend to change CESM part, 
only like to change exchange fields to SCHISM, like Hs, orbital, stoke's drift.... , 
please follow @DeniseWorthen suggestion, skip change within CESMCOUPLED .

sbanihash: @danishyo since this is your PR, what you will need to do is make these changes, commit and push them to your PR branch (which seems to be the dev/ufs-weather-model-branch in your fork) and then I will move forward with the review and testing for the PR. Thank you in advance.

DeniseWorthen:
@danishyo The fact that you see the decomposition reflected on the boundary gives me pause. I would expect if it were a simple issue of mapsta being incorrect, you might see garbage (or zero), but you're seeing something that looks like a decomposition. 

I don't know how much debugging you've done for this issue in WW3 itself, but I reminded of this issue I found when adding the unstructured mesh to the cap

https://github.com/ufs-community/ufs-weather-model/issues/1556#issuecomment-1458564534

In that case, the exported field was corrupted on the last PE, but only on the restart frequency. 

I suspect that something more fundamental may be occurring inside of WW3 it it is manifesting itself in your case.

danishyo:
> @danishyo The fact that you see the decomposition reflected on the boundary gives me pause. I would expect if it were a simple issue of mapsta being incorrect, you might see garbage (or zero), but you're seeing something that looks like a decomposition.
> 
> I don't know how much debugging you've done for this issue in WW3 itself, but I reminded of this issue I found when adding the unstructured mesh to the cap
> 
> [ufs-community/ufs-weather-model#1556 (comment)](https://github.com/ufs-community/ufs-weather-model/issues/1556#issuecomment-1458564534)
> 
> In that case, the exported field was corrupted on the last PE, but only on the restart frequency.
> 
> I suspect that something more fundamental may be occurring inside of WW3 it it is manifesting itself in your case.

Hi @DeniseWorthen 
I believe this is only related the code to derive orbital. I compare output for both WWM & WW3 (with update PR code), they are similar like the following figure.
圖片

Anyway, I will revert change within CESM block then push.

DeniseWorthen:
@danishyo For the current UFS cases using the unstructured mesh, we do not export some of the fields which you are using in the coastal app. So I never did any debugging of possible issues inside of WW3 w/ those fields. 

I don't doubt that you see that your fix resolves the issue---but I'm not 100% sure that the issue you are seeing isn't from something being done incorrectly inside of WW3. How else to explain that what you see is evidence of the decomposition of WW3 in the exported fields?

model/src/wav_import_export.F90
danishyo: @sbanihash Thanks for the explanation, I already revert the change within CESM block and push.

danishyo:
> @danishyo For the current UFS cases using the unstructured mesh, we do not export some of the fields which you are using in the coastal app. So I never did any debugging of possible issues inside of WW3 w/ those fields.
> 
> I don't doubt that you see that your fix resolves the issue---but I'm not 100% sure that the issue you are seeing isn't from something being done incorrectly inside of WW3. How else to explain that what you see is evidence of the decomposition of WW3 in the exported fields?

@DeniseWorthen I checked all exchange fields between SCHISM & WW3, orbital velocity is the only one showing this, others like Hs, Tm01... are not. So I just guessing this may not related.

model/src/wav_import_export.F90
```diff
@@ -1183,8 +1183,8 @@ subroutine CalcRoughl ( wrln)
       ix = mapsf(isea,1)
       iy = mapsf(isea,2)
       if ( firstCall ) then
-        if(( runtype == 'initial'  .and.     mapsta(iy,ix)  == 1 ) .or. &
-             ( runtype == 'continue' .and. abs(mapsta(iy,ix)) == 1 )) then
+        if(( runtype == 'initial'  .and.     mapsta(iy,ix)  /= 0 ) .or. &
```
DeniseWorthen: @danishyo Did you intend to change this block? I don't remember that coastal app exports Z0 and this is w/in code that UFS global mode uses.

danishyo: No, I didn't, I will revert this one as well.

DeniseWorthen:
> > @danishyo For the current UFS cases using the unstructured mesh, we do not export some of the fields which you are using in the coastal app. So I never did any debugging of possible issues inside of WW3 w/ those fields.
> > I don't doubt that you see that your fix resolves the issue---but I'm not 100% sure that the issue you are seeing isn't from something being done incorrectly inside of WW3. How else to explain that what you see is evidence of the decomposition of WW3 in the exported fields?
> 
> @DeniseWorthen I checked all exchange fields between SCHISM & WW3, orbital velocity is the only one showing this, others like Hs, Tm01... are not. So I just guessing this may not related.

I've had enough experience adding features and debugging in WW3 that when I see something like a decomposition field being expressed in the output field, it is difficult for me not be suspicious. The fact that you're seeing it in some things and not others doesn't really convince me. 

However, UFS is not making use of these fields that coastal is using currently, so if you restrict your changes to only coastal-app exported fields, I'm ok with your changes. I just would not be at all surprised to find something deeper in WW3 is actually causing a decomposition to appear in an exported field.

danishyo:
Thanks for the explanation, I just revert the changes and keep changes with necessary fields.

model/src/wav_import_export.F90
```diff
@@ -1492,8 +1492,8 @@ subroutine CalcStokes(a, us, vs, fval)
             abx = abx + a(ith,ik,jsea)*ecos(ith)
             aby = aby + a(ith,ik,jsea)*esin(ith)
           end do
-          kd = max ( 0.001 , wn(ik,isea) * dw(isea) )
-          if (kd .lt. 6.0) then
+          kd = max ( 1.0e-7 , wn(ik,isea) * dw(isea) ) !0.001
```
JessicaMeixner-NOAA, line 1495: Why are there changes to this calculation?  This appears different than the mapsta issue updated elsewhere but I don't see anything about it in the PR description.

JessicaMeixner-NOAA (commented):
I'm wondering if these should be if mapsta(iy,ix) > 0 as negative values indicate that a point is currently taken out of computation.  https://github.com/NOAA-EMC/WW3/blob/develop/model/src/w3gridmd.F90#L219-L247 

Also I'm not sure what prompted the changes in the calcstokes computations?

model/src/wav_import_export.F90
danishyo: I borrow routine coefficients from https://github.com/NOAA-EMC/WW3/blob/develop/model/src/wmesmfmd.F90#L7315
For mapsta, I am OK to change ">0" if this includes ocean and boundary points.

DeniseWorthen: The code for the ussco matches (or should) what is currently in w3iogomd.F90

JessicaMeixner-NOAA: @danishyo  - Can you let us know how your testing goes for ">0" this should be sufficient. 

I would think this calculation should match what is in w3iogo.  I'd have to look at the differences between that and wmesmf as to why that would have been diverged.   Let me know if you need help investigating this further @danishyo  or if you will take the lead on that.

JessicaMeixner-NOAA: @danishyo were you ever able to run those tests?

danishyo: > @danishyo were you ever able to run those tests?

I didn't do "mapsta>0" test, but base on info from manual, I believe this should fix the issue as well.

JessicaMeixner-NOAA: Let us know when you've been able to run those tests and confirm if that works or not.  

We also have outstanding issues about this part of the code change and why it's diverged, etc.   

This PR is waiting on those updates to move forward.

Issue #1580, opened 2026-03-12
IMPTOTAL solver stalling with REF1


leahcicon:
@aronroland @aliabdolali 
Description:

I believe I have come across a bug with the implicit solver (IMPTOTAL) using JGS_BLOCK_GAUSS_SEIDEL = T, and JGS_TERMINATE_DIFFERENCE = T with REF1 switch for unstructured grids. The issue that I am seeing is that the solver is not exiting early with TERMINATE_DIFFERENCE and the solver is hitting the maximum number of iterations. The proportion of converged nodes, prop_conv, is stalling at ~ 16% after ~70 iterations, so the JGS_TERMINATE_DIFFERENCE condition is never met. i.e. (prop_conv .le. B_JGS_PMIN). I believe I have traced this to PDLIB_JACOBI_GAUSS_SEIDEL_BLOCK in w3profsmd_pdlib.F90 and how the convergence metric is computed and with REF1 restoring reflected energy.

In w3profsmd_pdlib.F90 the convergence metric is:

`DiffNew = abs(sum(ACLOC - eSum)) / Sum_new`

Where:
ACLOC = previous VA
eSum = newly computed value

But with REF1 active:

- eSum is computed
- eSum is assigned to VA (VA = eSum)
 - Then parts of VA are overwritten with VAOLD (see code excerpt below).

Inside solver loop of PDLIB_JACOBI_GAUSS_SEIDEL_BLOCK:
```
#ifdef W3_REF1
  DO IK=1,NK
    DO ITH=1,NTH
      ISP  = ITH + (IK-1)*NTH
      IF (REFPARS(3) .LT. 0.5 .AND. IOBPD_LOC(ITH,IP) .EQ. 0 .AND. IOBPA_LOC(IP) .EQ. 0) THEN
        VA(ISP,IP) = VAOLD(ISP,IP) * IOBDP_LOC(IP)
      ENDIF
    ENDDO
  ENDDO
#endif
```
- For the next iteration, ACLOC includes those restored values. So the solver tries to move toward a solution and REF1 keeps snapping some components back. The solver reaches a state where the interior converges, the reflected boundary bins never change and so the residual stops decreasing. Which means the prop_conv stalls, approximately equal to the fraction of nodes affected by REF1. This is because the residual in the convergence metric is |ACLOC - eSum| where ACLOC has the restored values from the previous iteration but eSum does not.

Expected behavior:
The solver should be exiting early around ~70 iterations.

Possible solution:

[...]

PR #1608, opened 2026-06-24
Fix convergence stall in IMPTOTAL solver with REF1 and JGS_TERMINATE_DIFFERENCE on unstructured grids

leahcicon:
# Pull Request Summary
Fix convergence stall in IMPTOTAL solver with REF1 and JGS_TERMINATE_DIFFERENCE on unstructured grids.

## Description
This PR fixes a bug in PDLIB_JACOBI_GAUSS_SEIDEL_BLOCK in w3profsmd_pdlib.F90 where the implicit solver (IMPTOTAL) fails to exit early when using JGS_BLOCK_GAUSS_SEIDEL = T, JGS_TERMINATE_DIFFERENCE = T, and the REF1 switch on unstructured grids. Instead of converging early, the solver stalls at approximately 16% converged nodes and runs to the maximum number of iterations. The number of converged nodes will change depending on the amount of grid nodes affected by REF1.

Root Cause:
The convergence metric was computed as:
```
DiffNew = abs(sum(ACLOC - eSum)) / Sum_new
```
Where ACLOC is the previous spectrum and eSum is the newly computed value. However, when REF1 is active, after eSum is assigned to VA, certain bins of VA are overwritten with VAOLD:

```
#ifdef W3_REF1
  DO IK=1,NK
    DO ITH=1,NTH
      ISP  = ITH + (IK-1)*NTH
      IF (REFPARS(3) .LT. 0.5 .AND. IOBPD_LOC(ITH,IP) .EQ. 0 .AND. IOBPA_LOC(IP) .EQ. 0) THEN
        VA(ISP,IP) = VAOLD(ISP,IP) * IOBDP_LOC(IP)
      ENDIF
    ENDDO
  ENDDO
#endif
```

On the next iteration, ACLOC picks up these restored values. The result is that the interior nodes converge normally but the REF1 boundary bins never change between iterations. The residual |ACLOC - eSum| remains nonzero for those bins because eSum does not include the REF1 overwrite. prop_conv stalls at approximately the fraction of non-REF1 nodes and the JGS_TERMINATE_DIFFERENCE condition is never satisfied, causing the solver to run to maximum iterations.

Fix:
The convergence metric is changed to compare against VA after the REF1 overwrite has been applied, rather than eSum before it:
```
IF (B_JGS_TERMINATE_DIFFERENCE) THEN
  Sum_New = sum(VA(:,IP))
  if (Sum_new .gt. 0.d0) then
    DiffNew = abs(sum(ACLOC-VA(:,IP)))/Sum_new
    p_is_converged = DiffNew
  else
    p_is_converged = zero
  endif
  IF (p_is_converged .lt. B_JGS_DIFF_THR .and. nbiter .gt. 1) then
    is_converged   = is_converged + 1
    lconverged(ip) = .true.
  ELSE
    lconverged(ip) = .false.
  ENDIF
END IF
```

With this fix, for REF1 bins ACLOC - VA = 0, so those bins no longer pollute the convergence metric and the solver exits normally once the interior nodes converge.

Labels: bug
Answer changes: No change in answers is expected, the changes just allow for the early exit to happen.

### Issue(s) addressed
- fixes issue #1580

### Commit Message
Fix convergence stall in IMPTOTAL solver with REF1 and JGS_TERMINATE_DIFFERENCE on unstructured grids. The convergence metric in PDLIB_JACOBI_GAUSS_SEIDEL_BLOCK in w3profsmd_pdlib.F90 now compares against VA after REF1 overwrite is applied rather than eSum before it, preventing REF1 boundary bins from polluting the residual and stalling prop_conv.

### Check list  

### Testing

How were these changes tested? Preliminary runs with the fix applied show the solver exiting early around ~70 iterations as expected, rather than running to the maximum iteration count.
Are the changes covered by regression tests? Not currently. A regression test covering IMPTOTAL with REF1 on an unstructured grid with JGS_TERMINATE_DIFFERENCE = T would be needed to fully cover this fix.
Have the matrix regression tests been run? No
Expected changes in regression test output: No changes expected in existing regression tests since this code path is only active with JGS_TERMINATE_DIFFERENCE = T and REF1 on unstructured grids.
matrix.comp output: Not available.


mingchen-NOAA:
Matrix Comparison tests were performed on Ursa-Intel.

Known non-identical cases
```
mww3_test_03/./work_PR3_UNO_MPI_d2                     (14 files differ)
mww3_test_03/./work_PR3_UQ_MPI_e                     (1 files differ)
mww3_test_03/./work_PR3_UNO_MPI_e                     (1 files differ)
mww3_test_03/./work_PR1_MPI_d2                     (14 files differ)
mww3_test_03/./work_PR3_UQ_MPI_e_c                     (1 files differ)
mww3_test_03/./work_PR2_UNO_MPI_d2                     (17 files differ)
mww3_test_03/./work_PR3_UQ_MPI_d2_c                     (16 files differ)
mww3_test_03/./work_PR2_UNO_MPI_e                     (1 files differ)
mww3_test_03/./work_PR3_UQ_MPI_d2                     (16 files differ)
mww3_test_03/./work_PR1_MPI_e                     (1 files differ)
mww3_test_03/./work_PR3_UNO_MPI_d2_c                     (15 files differ)
mww3_test_03/./work_PR3_UNO_MPI_e_c                     (1 files differ)
mww3_test_09/./work_MPI_ASCII                     (0 files differ)
ww3_tp2.6/./work_ST4_ASCII                     (0 files differ)
ww3_ufs1.3/./work_a                     (3 files differ)
```

Unexpected differs
```
ww3_tp2.17/./work_ice_restart                     (34 files differ)
ww3_tp2.17/./work_c                     (8 files differ)
ww3_tp2.17/./work_deep                     (44 files differ)
ww3_tp2.17/./work_iceB                     (44 files differ)
ww3_tp2.17/./work_deepB                     (44 files differ)
ww3_tp2.17/./work_mc                     (6 files differ)
ww3_tp2.17/./work_ice                     (44 files differ)
ww3_tp2.17/./work_mc1                     (6 files differ)
ww3_tp2.17/./work_deep_restart                     (34 files differ)
ww3_tp2.19/./work_1A_a                     (9 files differ)
ww3_tp2.19/./work_1C_a                     (9 files differ)
ww3_tp2.19/./work_1B_a                     (9 files differ)
ww3_tp2.21/./work_b                     (6 files differ)
ww3_tp2.21/./work_b_metis                     (6 files differ)
ww3_tp2.21/./work_mb                     (7 files differ)
ww3_tp2.6/./work_pdlib                     (11 files differ)
ww3_ufs1.1/./work_unstr_c                     (365 files differ)
ww3_ufs1.1/./work_unstr_d                     (365 files differ)
```
[matrixCompFull.txt](https://github.com/user-attachments/files/29891738/matrixCompFull.txt)
[matrixCompSummary.txt](https://github.com/user-attachments/files/29891739/matrixCompSummary.txt)

mingchen-NOAA:
@leahcicon In your PR description, you mentioned 

'No changes expected in existing regression tests since this code path is only active with JGS_TERMINATE_DIFFERENCE = T and REF1 on unstructured grids.' 

However, our regression testing shows a number of unexpected result changes. Could you help explain why these changes are occurring and confirm whether they are expected?

leahcicon:
Sorry the late reply, I've been away on vacation. 

I am less familiar with the regtests so just to clarify the meaning of the differences you've provided taking the `mww3_test_03./work_PR3_UNO_MPI_d2` as an example. This regtest uses the switches in `switch_PR3_UNO_MPI` and runs ww3_multi using `ww3_multi_grdset_d2.nml` and the low1 low2 low3 hgh1 hgh2 and hgh3 grids and their corresponding namelist files Ex. `namelists_low1.nml`? If that's correct in this case I don't understand how the results could differ because the model doesn't use the implicit solving scheme (IMPTOTAL=T) amongst other things.

I traced back the settings which could result in differences based on the changes which are only in `w3profsmd_pdlib.F90`.

- Changes are made in `PDLIB_JACOBI_GAUSS_SEIDEL_BLOCK` function in `w3profsmd_pdlib.F90` and are only applied or implemented when `B_JGS_TERMINATE_DIFFERENCE =T`.
- `PDLIB_JACOBI_GAUSS_SEIDEL_BLOCK` is only run if `B_JGS_USE_JACOBI =T` in the `PDLIB_W3XYPUG_BLOCK_IMPLICIT` function.
``` 
IF (B_JGS_USE_JACOBI) THEN
        CALL PDLIB_JACOBI_GAUSS_SEIDEL_BLOCK(IMOD, FACX, FACY, DTG, VGX, VGY, LCALC)
        RETURN
```

-  `PDLIB_W3XYPUG_BLOCK_IMPLICIT` is called in `w3wavemd.F90` if `IMPTOTAL = T` and with PDLIB switch.

Therefore, I believe these changes could only result in differences when using PDLIB switch, with `IMPTOTAL = T`, `JGS_USE_JACOBI = T`, and using `B_JGS_TERMINATE_DIFFERENCE =T`. So I can't explain when the output differs for regtests when this isn't the case.

But for `regtest ww3_ufs1.1/./work_unstr_c` it has the solver settings to actually run the changes (I wasn't aware of this regtest). However, this set up uses REF0 so the stalling of the solver won't actually occur because the stalling only happens with REF1. But I am interested in the fact that with REF0 there seems to be an effect on results, which I don't expect.

mingchen-NOAA:
@leahcicon We have a known 'non-identical' list in the PR template 

- Please indicate the expected changes in the regression test output, (Note the [list](https://github.com/NOAA-EMC/WW3/wiki/How-to-use-matrix.comp-to-compare-regtests-with-develop#4-look-at-results) of known non-identical tests.)

For the cases on that list, the results are not reproducible, so you do not need to worry about those differences. 
However, the following cases are not on the known non-identical list:

```
ww3_tp2.17/./work_ice_restart                     (34 files differ)
ww3_tp2.17/./work_c                     (8 files differ)
ww3_tp2.17/./work_deep                     (44 files differ)
ww3_tp2.17/./work_iceB                     (44 files differ)
ww3_tp2.17/./work_deepB                     (44 files differ)
ww3_tp2.17/./work_mc                     (6 files differ)
ww3_tp2.17/./work_ice                     (44 files differ)
ww3_tp2.17/./work_mc1                     (6 files differ)
ww3_tp2.17/./work_deep_restart                     (34 files differ)
ww3_tp2.19/./work_1A_a                     (9 files differ)
ww3_tp2.19/./work_1C_a                     (9 files differ)
ww3_tp2.19/./work_1B_a                     (9 files differ)
ww3_tp2.21/./work_b                     (6 files differ)
ww3_tp2.21/./work_b_metis                     (6 files differ)
ww3_tp2.21/./work_mb                     (7 files differ)
ww3_tp2.6/./work_pdlib                     (11 files differ)
ww3_ufs1.1/./work_unstr_c                     (365 files differ)
ww3_ufs1.1/./work_unstr_d                     (365 files differ)
```

These differences appear to be introduced by this PR. We need to confirm that they are expected and understood before merging. We cannot merge a PR that unintentionally changes regression test results and potentially impacts other users.

leahcicon:
Okay I had to do a little investigating but I believe I have a good understanding of why some regtests show differences. I was under the impression the stalling convergence bug was only affecting REF1 but its actually present in REF0. With REF0 if JGS_DIFF_THR is small enough ~1E-8 the solver will stall in the same manner as REF1 but its due to a precision difference issue when computing DiffNew where eSum (double precision) compared against ACLOC (read from single-precision VA) opposed to the original bug I described.

I now have an equivalent but more generalized fix to the stalling convergence issue. There will be differences in regtest with REF1 and REF0 with JGS_BLOCK_GAUSS_SEIDEL=T and  JGS_DIFF_THR=1.E-8 because the model is now actually exiting early from JGS_TERMINATE_DIFFERENCE=T instead of hitting JGS_MAXITER.

What is the best way to approach this? Should I submit an entirely new merge request with the alternate bug fix?

mingchen-NOAA:
Thanks for investigating this and for the detailed explanation.

If you can incorporate the more generalized fix into the current PR while keeping the scope clear, I'm happy for you to update this PR and I'll retest it.

If you think the new approach substantially changes the scope or direction of the current PR, it would also be perfectly reasonable to close this PR and open a new one with the revised fix.

Either approach works for me. The important thing is that the PR description clearly explains the issue being addressed, the rationale for the chosen solution, and the expected regression test differences.

leahcicon:
I submitted a new pull request.

